PR #31, opened 2025-01-30
Added confirmation dialog when user clicks the delete button.

lizlooney:
Added confirmation dialog when user clicks the delete button.
Changed position of tooltips for New Workspace and Rename buttons because the left side was clipped.


alan412 (approved):
LGTM

PR #45, opened 2025-02-14
Added mechanism modules to the app

lizlooney:
Added mechanisms to the app (new mechanism, show mechanism in the tree, rename/copy mechanism, etc).

Set the title of the workspace name modal bases on what is being done (new workspace, rename workspace, copy workspace). Set the title of the module name modal bases on what is being done (new workspace, rename workspace, copy workspace).

Fixed a problem where the module type wasn't being passed to the python generator when it generates the python code for the UI.

When a new module is selected and we are updating the blocks editor area, close the toolbox flyout if it is open.

When saving a module to storage, include the module type. Improved the code that writes and reads modules (Workspaces, Mechanisms, and OpModes) to/from storage.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,14 +19,16 @@ import {
 import type { InputRef } from 'antd';
 import type { TreeDataNode, TreeProps } from 'antd';
 import {
+  AppstoreAddOutlined as MechanismAddOutlined,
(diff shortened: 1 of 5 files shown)


alan412 (approved):
I think we should improve the icon for mechanism, but this looks like a good way to do this to me.

PR #48, opened 2025-02-15
Change the "Workspace" terminology to "Project"

### LINKED ISSUE #36: Change "Workspace" to "Project"
As decided on Feb 6, we are going to change the "Workspace" terminology to "Project" so it won't conflict with the idea of a Blockly Workspace....

lizlooney:
Change the "Workspace" terminology to "Project" so it won't conflict with the idea of a Blockly Workspace.

Fixes #36

### DIFF
--- a/src/storage/client_side_storage.ts
+++ b/src/storage/client_side_storage.ts
@@ -167,7 +167,7 @@ export function listModules(callback: ModulesCallback): void {
     if (cursor) {
       const value = cursor.value;
       const path = value.path;
-      const moduleType = _convertValueTypeToModuleType(value.type);
+      const moduleType = value.type;
       const module: commonStorage.Module = {
         modulePath: path,
         moduleType: moduleType,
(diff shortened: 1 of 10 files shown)


src/storage/client_side_storage.ts
```diff
@@ -167,78 +167,87 @@ export function listModules(callback: ModulesCallback): void {
     if (cursor) {
       const value = cursor.value;
       const path = value.path;
+      const moduleType = _convertValueTypeToModuleType(value.type);
```
alan412, line 170: Should this really be in the code since we really are only supporting Project?  (ie, there aren't legacy versions out in the wild)

src/storage/client_side_storage.ts
```diff
@@ -167,78 +167,87 @@ export function listModules(callback: ModulesCallback): void {
...
+      const sortedProjectNames = Object.keys(projects).sort();
+      sortedProjectNames.forEach((projectName) => {
+        modules.push(projects[projectName]);
       });
       callback(modules, '');
     }
   };
 }
 
+function _convertValueTypeToModuleType(valueType: string): string {
```
alan412, line 243: Not sure this should be here.

alan412 (changes requested):

src/storage/client_side_storage.ts
lizlooney: I removed it and added a note in the readme.

alan412 (approved):
LGTM

PR #51, opened 2025-02-21
Fixed code that triggers python regeneration.

lizlooney:


alan412 (approved):
Pulled down and verified

PR #55, opened 2025-02-22
Remove duplicate comment from methods

### LINKED ISSUE #25: Comment on method shows up twice
Adding a comment to a method block has it show up twice in the python code

alan412:
Fixes #25

### DIFF
--- a/src/blocks/mrc_class_method_def.ts
+++ b/src/blocks/mrc_class_method_def.ts
@@ -337,12 +337,13 @@ export const setup = function() {
   Blockly.Blocks[PARAM_CONTAINER_BLOCK_NAME] = METHOD_PARAM_CONTAINER;
 };
 
-import { Order, PythonGenerator } from 'blockly/python';
+import { Order } from 'blockly/python';
+import { ExtendedPythonGenerator } from '../editor/extended_python_generator';
 
 
 export const pythonFromBlock = function (
     block: ClassMethodDefBlock,
-    generator: PythonGenerator,
+    generator: ExtendedPythonGenerator,
 ) {
     const blocklyName = block.mrcPythonMethodName ? block.mrcPythonMethodName : block.getFieldValue('NAME');
 
@@ -378,6 +379,9 @@ export const pythonFromBlock = function (
         // After executing the function body, revisit this block for the return.
         xfix2 = xfix1;
     }
+    if(block.mrcPythonMethodName == '__init__'){
+        branch = generator.INDENT + "self.mechanisms = []\n" + branch;
+    }    
     if (returnValue) {
         returnValue = generator.INDENT + 'return ' + returnValue + '\n';
     } else if (!branch) {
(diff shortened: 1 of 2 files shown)


src/blocks/mrc_class_method_def.ts
```diff
@@ -337,12 +337,13 @@ export const setup = function() {
   Blockly.Blocks[PARAM_CONTAINER_BLOCK_NAME] = METHOD_PARAM_CONTAINER;
 };
 
-import { Order, PythonGenerator } from 'blockly/python';
+import { Order } from 'blockly/python';
```
lizlooney, line 340: Do these imports have to be here? Can they be at the top of the file? https://google.github.io/styleguide/tsguide.html#source-file-structure

src/editor/extended_python_generator.ts
```diff
@@ -31,6 +31,8 @@ import * as commonStorage from '../storage/common_storage';
 export class ExtendedPythonGenerator extends PythonGenerator {
   private currentModule: commonStorage.Module | null = null;
   private mapWorkspaceIdToExportedBlocks: { [key: string]: Block[] } = Object.create(null);
+  protected methods_: {[key: string]: string} = Object.create(null);
```
lizlooney, line 34: If this generator is used to generate code for multiple blocks workspaces (for example, the workspace in the UI as well as headless workspaces that we use for creating new modules), will this methods_ object get messed up?

lizlooney: At this point, I'm ok with this being a problem that we create an issue for it. Then we can think about what the best solution is. I'll create an issue and put in some notes for discussion.

lizlooney, line 34: Why protected instead of private?
Why the trailing _?

alan412: I almost always go for protected instead of private so it can be subclassed later.
The trailing _ was because I was copying the style from `definitions_`

src/blocks/mrc_class_method_def.ts
lizlooney: Sure.

src/editor/extended_python_generator.ts
lizlooney: I'm the opposite. I go for the most restrictive visibility first and then make it more visible only if necessary.
If it needs to be subclassed later, it can be changed from private to protected then.
https://google.github.io/styleguide/tsguide.html#visibility

alan412: While I agree fully about not making things public by default (as the style guide says), protected seems to be better than private.   Imagine the mess we would be in if Blockly had made everything private instead of protected.    (I'll admit that this tendency of mine is because I am almost always creating libraries or systems instead of applications.)

lizlooney: That makes sense for libraries where you (as author of the library) want to allow a class to be subclassed. In this case, it is not a library and within this application, we don't currently want to subclass ExtendedPythonGenerator.

I'm going to approve (and merge) this PR so we can keep making progress.

lizlooney (approved):

PR #61, opened 2025-02-22
Asynchronous storage API

### LINKED ISSUE #53: Use async for storage functions instead of callbacks

lizlooney:
Fixes #53


alan412 (commented):
This feels like such a major change that there should be some tests for it.   I'll try to run it through some manual testing tonight, but I wonder if we could create any automated tests for it.

alan412:
did some rough manual testing.   LGTM

PR #68, opened 2025-02-25
Make mrc_call_python_function work for calling methods in the same class.

### LINKED ISSUE #64: Make mrc_call_python_function work for calling methods in the same class.
Currently mrc_call_python_function assumes that the function being called is outside of the current class/module. It needs to be updated so it can call a method that is defined in the same class.

lizlooney:
Make mrc_call_python_function work for calling methods in the same class.

Fixes #64

### DIFF
--- a/src/blocks/mrc_call_python_function.ts
+++ b/src/blocks/mrc_call_python_function.ts
@@ -33,11 +33,13 @@ import { MRC_STYLE_FUNCTIONS } from '../themes/styles'
 
 export const BLOCK_NAME = 'mrc_call_python_function';
 
-const FUNCTION_KIND_MODULE = 'module';
-const FUNCTION_KIND_STATIC = 'static';
-const FUNCTION_KIND_CONSTRUCTOR = 'constructor';
-const FUNCTION_KIND_INSTANCE = 'instance';
-const FUNCTION_KIND_INSTANCE_WITHIN = 'instance_within';
+enum FunctionKind {
+  MODULE = 'module',
+  STATIC = 'static',
+  CONSTRUCTOR = 'constructor',
+  INSTANCE = 'instance',
+  INSTANCE_WITHIN = 'instance_within',
+}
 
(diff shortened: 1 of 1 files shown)


src/blocks/mrc_call_python_function.ts
```diff
@@ -127,6 +128,11 @@ const CALL_PYTHON_FUNCTION = {
           tooltip = 'Calls the function ' + className + '.' + functionName + '.';
           break;
         }
+        case FUNCTION_KIND_INSTANCE_WITHIN: {
+          const functionName = this.getFieldValue(pythonUtils.FIELD_FUNCTION_NAME);
+          tooltip = 'Calls the method ' + functionName + '.';
+          break;
+        }
         default:
           throw new Error('mrcVarKind must be "module", "static", "constructor", or "instance".')
```
alan412, line 137: Does this need to say or instance_within?

src/blocks/mrc_call_python_function.ts
```diff
@@ -37,6 +37,7 @@ const FUNCTION_KIND_MODULE = 'module';
 const FUNCTION_KIND_STATIC = 'static';
 const FUNCTION_KIND_CONSTRUCTOR = 'constructor';
 const FUNCTION_KIND_INSTANCE = 'instance';
+const FUNCTION_KIND_INSTANCE_WITHIN = 'instance_within';
```
alan412, line 40: should these be string enums so the compiler will only allow the ones we want?

src/blocks/mrc_call_python_function.ts
lizlooney: Good catch.

src/blocks/mrc_call_python_function.ts
lizlooney: Yes!

lizlooney:
Thanks for the review. PTAL (please take another look)

alan412:
LGTM

PR #79, opened 2025-02-28
Modified ExtendedPythonGenerator so it can safely be a singleton

### LINKED ISSUE #63: A single ExtendedPythonGenerator instance is used to generate python for multiple workspaces.
We either need to make sure that the single ExtendedPythonGenerator instance is being correctly used for multiple blockly workspaces, or we need to use multiple ExtendedPythonGenerator instances.

Right now there is a single ExtendedPythonGenerator instance, the extendedPythonGenerator variable that is declared in extended_python_generator.ts. But that generator is used to generate code for multiple blockly workspaces (for example, the workspace in the UI as well as headless blockly workspaces that we use for creating new modules, etc).

We have a method in ExtendedPythonGenerator called **setCurrentModule** that is called immediately before we call extendedPythonGenerator.workspaceToCode. This is done in the following places:

1. In App.tsx, for generating the code that is displayed in the Python Code panel that is to the right of the visible Blockly Workspace.
2. In editor/editor.ts, for generating the code that is saved to storage.
3. In storage/common_storage.ts, for generating code when we create a new Project, Mechanism, or OpMode, or when we upload a Project.

We have another method in ExtendedPythonGenerator called **getExportedBlocks** that is called immediately after workspaceToCode after we call extendedPythonGenerator.workspaceToCode. This is done in the same places listed above, except not in App.tsx since it is not needed there.

In ExtendedPythonGenerator.**init** we collect information about the functions and variables that have been declared so we can produce the exported blocks.

lizlooney:
Fixed #63

Added generator_context.ts with class GeneratorContext, which holds information needed during and after python code generation:
- The module (so the ExtendedPythonGenerator can get the class and parent class names)
- Exported blocks (these are produced by the ExtendedPythonGenerator and are added to the file after code generation is finished).
- The class method names (maps from the block's NAME field to the actual python method name)
- Method getClassName returns the class name as PascalCase.

Modified ExtendedPythonGenerator so it can safely be a singleton.
- Removed fields currentModule and mapWorkspaceIdToExportedBlocks and added context (type GeneratorContext).
- Renamed methods_ field to classMethods. This field can stay in ExtendedPythonGenerator because it is only needed during workspaceToCode execution. It is not needed after workspaceToCode has finished.
- Added a field name workspace for the Blockly Workspace because it is needed in the finish method, but it is not a parameter.
- Removed init method override.
- Added workspaceToCode overload with additional parameter context (type GeneratorContext).
- Renamed method addMethod to addClassMethodDefinition and added NAME field value as a parameter.
- Added finish method override. --  Here we still produce the code, which is the class definition line and the class method definitions. --  By setting the code parameter to the code we want (the class definition) and passing it to super.finish, we can just call super.finish normally. --  Produce the exported blocks.
- Moved setCurrentModule (renamed to setModule), getExportedBlocks, and classParentFromModuleType (renamed to getClassParent) to GeneratorContext.

Updated the App to create a GeneratorContext and pass it to workspaceToCode. Updated common_storage to create a GeneratorContext and pass it to workspaceToCode. Updated Editor to create a GeneratorContext and pass it to workspaceToCode.

Modified code generator for mrc_class_method_def blocks to return empty string.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -253,7 +253,7 @@ const App: React.FC = () => {
   const [deleteTooltip, setDeleteTooltip] = useState('Delete');
   const blocklyComponent = useRef<BlocklyComponentType | null>(null);
   const [triggerPythonRegeneration, setTriggerPythonRegeneration] = useState(0);
-  const generatorContext = useRef(createGeneratorContext());
+  const generatorContext = useRef<GeneratorContext | null>(null);
   const blocksEditor = useRef<editor.Editor | null>(null);
   const [generatedCode, setGeneratedCode] = useState('');
   const [newProjectNameModalPurpose, setNewProjectNameModalPurpose] = useState('');
@@ -534,7 +534,8 @@ const App: React.FC = () => {
       blocklyWorkspace.addChangeListener(mutatorOpenListener);
       blocklyWorkspace.addChangeListener(handleBlocksChanged);
     }
-    blocksEditor.current = new editor.Editor(blocklyWorkspace, storage);
+    generatorContext.current = createGeneratorContext();
+    blocksEditor.current = new editor.Editor(blocklyWorkspace, generatorContext.current, storage);
   }, [blocklyComponent, storage]);
 
   const handleBlocksChanged = (event: Blockly.Events.Abstract) => {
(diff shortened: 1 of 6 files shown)


alan412 (approved):
LGTM and verified it works on my system

PR #82, opened 2025-02-28
Fixes "Blocks have been modified" dialog appears twice bug

### LINKED ISSUE #81: "Blocks have been modified" dialog appears twice
If you modify the blocks and then click on a different module in the tree, the "Blocks have been modified! Press ok to save and continue" dialog appears. If you click OK, it appears again.

It should not re-appear after you click OK.

lizlooney:
Use await when calling saveBlocks.

Fixes #81


alan412 (approved):
Tested and it works for me.

PR #84, opened 2025-03-01
Handle method names that are python reserved words.

### LINKED ISSUE #73: Handle method names that are python reserved words
If the user creates a mrc_class_method_def block with the name "raise", the generated python code has a different name. The block's mrcPythonMethodName field should have this different name so we can create a mrc_call_python_function block correctly.

lizlooney:
The code generated for blocks that call methods defined in the same module match the method definition blocks. For example, if you create a method named "raise" and you drop a call block for it on the same blockly workspace, the python code for both the definition and the call will be "raise2".

Added known issue to readme: renaming a method definition doesn't update call blocks.

Fixes #73

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,3 +16,4 @@ WARNING! This is not ready for use and is under heavy development of basic featu
 2. Mechanisms aren't limited to only Robot or Mechanism class
 3. No way to specify whether an opmode is auto or teleop
 4. Since we changed the "Workspace" terminology to "Project", existing Workspaces are no longer supported. They can be deleted via the browser's Developer Tools - Application tab.
+5. When a method definition block's name or parameters are changed, the call blocks don't update automatically.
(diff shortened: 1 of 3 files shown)


alan412 (commented):
This doesn't seem like the right way to solve this problem.   It feels like it should all be done in mrc_class_method_def using .mrcPythonMethodName.  That way the same code we will need for when a user renames a method can be used for when they put in a reserved word

alan412:
Although i did test it and it does work.

lizlooney:
> It feels like it should all be done in mrc_class_method_def using .mrcPythonMethodName.

I agree! That would be the best solution!

I tried to set .mrcPythonMethodName to funcName, like this:
Screenshot 2025-03-03 at 10 05 03 PM

But, that made it so the python method name stayed that name, even if the user changed the method name on the block. I looked for where it handles the event that happens when the user edits the method name on the blocks (so I could set .mrcPythonMethodName to ''), but I couldn't find a handler for that. Is there already an event handler for that?

lizlooney:
I made some more progress on this by looking at how blockly does their procedure blocks.

lizlooney:
I'm closing this PR. I'll make a new one with a better implementation.

PR #88, opened 2025-03-05
Handle method names that are python reserved words.

### LINKED ISSUE #73: Handle method names that are python reserved words

lizlooney:
Handle method names that are python reserved words.
Fixes #73   

Update mrc_call_python_function blocks in the same blockly workspace when the method name on a mrc_class_method_def block is changed.
(This does not close 71 because does not update callers if a parameter or the return type changes.

### DIFF
--- a/src/blocks/mrc_call_python_function.ts
+++ b/src/blocks/mrc_call_python_function.ts
@@ -48,7 +48,7 @@ export type FunctionArg = {
   type: string,
 };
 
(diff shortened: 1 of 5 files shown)


alan412 (approved):
I like it!   Much cleaner than before and it works like a champ!!

PR #93, opened 2025-03-08
Update parameters for call block when method args are changed.

### LINKED ISSUE #71: Update mrc_call_python_function blocks in the same blockly workspace if a mrc_class_method_def block is changed
We need to handle changes the the name, parameters, and return type.

lizlooney:
Update parameters for call blocks in the same blockly workspace if a method def block is changed.

Fixes #71


alan412 (approved):
Verified both the renaming, adding parameters and removing parameters

PR #94, opened 2025-03-08
Prepare for consolidating UI strings

alan412:
Before I do the rest of this, I want to make sure you are ok with the approach.

Instead of just pulling them into a file, I went ahead and pulled in i18n support (even though we will start with just English.)


src/i18n/config.ts
```diff
@@ -0,0 +1,52 @@
...
+    // Specifies the default language (locale) used
+    // when a user visits our site for the first time.
+    lng: "en",
+
+    // Fallback locale used when a translation is missing.
+    fallbackLng: "en",
+
+    // Normally, we want `escapeValue: true` as it ensures that i18next escapes any code in
+    // translation messages, safeguarding against XSS (cross-site scripting) attacks. However,
+    // React does this escaping itself, so we turn it off in i18next.
```
lizlooney, line 46: How do you know that react does the escaping itself?

lizlooney (commented):
This looks great! Keep going!

src/i18n/config.ts
alan412: That was a comment that came from the web page that had the sample code I liberally borrowed from so I guess I am trusting that is true...

alan412: Ok - I did a search and found on https://react.i18next.com/latest/using-with-hooks a more trusted version.

Looking up React, you can't turn it off but if you need to put in an unescaped string you do: `dangerouslySetInnerHTML`

PR #97, opened 2025-03-10
Quick pass at sample components

alan412:
Here is a quick - what the base class might look like and a sample one.  I think by requiring type hints that we might be able to get away with no blocks specific information needed.


external_samples/component.py
```diff
@@ -0,0 +1,57 @@
...
+    def stop(self) -> None:
+        pass
+
+    # any reset required (if any) at the beginning of each opmode
+    # This might remove any registered callbacks
+    @abstractmethod
+    def reset(self) -> None:
+        pass
+
+    # This returns a list (can be empty, one, or multipe) of the ports this connects to
```
lizlooney, line 48: multipe -> multiple

external_samples/component.py
```diff
@@ -0,0 +1,57 @@
...
+    # of the PortType enumeration
+    @abstractmethod
+    def get_connection_port_type(self) -> list[PortType]:
+        pass
+
+    # This is called periodically when an opmode is running
+    @abstractmethod
+    def periodic(self) -> None:
```
lizlooney, line 56: Interesting. Can you give an example (like what you have in rev_touch_sensor.py) of what a component subclass might do?

external_samples/rev_touch_sensor.py
```diff
@@ -0,0 +1,38 @@
+from component import Component, PortType
+from _collections_abc import Callable
```
lizlooney, line 2: from collections.abc import Callable

external_samples/rev_touch_sensor.py
```diff
@@ -0,0 +1,38 @@
...
+    def get_version(self) -> tuple[int, int, str]: 
+        return (1, 0, "")
+    def stop(self) -> None:
+        pass
+    def reset(self) -> None:
+        pass
+    def get_connection_port_type(self) -> list[PortType]:
+        return [PortType.SMART_IO_PORT]
+    def periodic(self) -> None:
+        # This would poll the hardware and see if it needs to fire any events
```
lizlooney, line 23: I think this comment would be helpful in component.py.

lizlooney (commented):
Looks good. I made a few comments.

external_samples/component.py
```diff
@@ -0,0 +1,68 @@
+from abc import ABC, abstractmethod
```
lizlooney, line 1: ABC isn't used

external_samples/component.py
```diff
@@ -0,0 +1,68 @@
...
+    @abstractmethod
+    def get_part_number(self) -> str:  
+        pass
+    # This is the URL of the component
+    @abstractmethod
+    def get_url(self) -> str:  
+        pass
+    # This is the version of the software (returned as a (major, minor, revision) tuple where 
+    # major and minor are positive integers
+    # revision is an optional string
```
lizlooney, line 42: Does optional mean that the returned tuple can have just 2 items? Or that the last item might be an empty string or None?

external_samples/component.py
```diff
@@ -0,0 +1,68 @@
...
+    @abstractmethod
+    def get_version(self) -> tuple[int, int, str]: 
+        pass
+    
+    # This stops all movement (if any) for the component
+    @abstractmethod
+    def stop(self) -> None:
+        pass
+
+    # any reset required (if any) at the beginning of each opmode
```
lizlooney, line 52: # any -> # Performs any

external_samples/rev_touch_sensor.py
```diff
@@ -0,0 +1,53 @@
+from component import Component, PortType, InvalidPortException, EmptyCallable
+
+class RevTouchSensor(Component):
+    def __init__(self, ports : list[tuple[PortType, int]]):
+        self.is_pressed = None
```
lizlooney, line 5: It seems like is_pressed should be a boolean, so the value should be False instead of None.

external_samples/rev_touch_sensor.py
```diff
@@ -0,0 +1,53 @@
...
+        # here read hardware to get the current value of the sensor and set self.is_pressed
+        pass
+
+    def is_pressed(self) -> bool:
+        '''Returns if the touch sensor is pressed or not'''
+        return self.is_pressed
+    
+    # Events
+    def register_when_pressed(self, callback: EmptyCallable) -> None:
+        '''Event when touch sensor is first pressed'''
```
lizlooney, line 47: Remove the word first. Otherwise it sounds like the event will only fire once, even if you press and release the touch sensor many times.

external_samples/rev_touch_sensor.py
```diff
@@ -0,0 +1,53 @@
...
+        self.pressed_callback = callback
+
+
+    def register_when_released(self, callback: EmptyCallable) -> None:
+        '''Event when touch sensor is first released'''
```
lizlooney, line 52: Remove the word first.

external_samples/servo.py
```diff
@@ -0,0 +1,42 @@
+from component import Component, PortType, InvalidPortException
+from collections.abc import Callable
```
lizlooney, line 2: Callable is not used.

external_samples/color_range_sensor.py
```diff
@@ -0,0 +1,63 @@
+from component import Component, PortType, InvalidPortException
+from collections.abc import Callable, Protocol
```
lizlooney, line 2: Callable is not used.

external_samples/servo.py
```diff
@@ -0,0 +1,42 @@
...
+
+class Servo(Component):    
+    # Required methods
+    def __init__(self, ports : list[tuple[PortType, int]]):
+        self.is_pressed = None
```
lizlooney, line 7: Remove this line.

external_samples/smart_motor.py
```diff
@@ -0,0 +1,51 @@
+from component import Component, PortType, InvalidPortException
+
+class SmartMotor(Component):    
+    # Required methods
+    def __init__(self, ports : list[tuple[PortType, int]]):
+        self.is_pressed = None
```
lizlooney, line 6: Remove this line.

lizlooney (changes requested):

external_samples/color_range_sensor.py
```diff
@@ -0,0 +1,63 @@
...
+        pass
+    def reset(self) -> None:
+        pass
+    def get_connection_port_type(self) -> list[PortType]:
+        return [PortType.I2C_PORT]
+    def periodic(self) -> None:
+        pass
+
+    # Component specific methods
+    def get_color_rgb(self) -> list[int, int, int]:
```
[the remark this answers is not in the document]
alan412: I wonder if it should be a tuple instead.  It is important that it return the three parts, not a list of any length.

external_samples/component.py
```diff
@@ -1,6 +1,10 @@
 from abc import ABC, abstractmethod
 from enum import Enum
+from collections.abc import Callable, Protocol
 
+class EmptyCallable(Protocol):
```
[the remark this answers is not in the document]
alan412: It is my understanding of what you have to do to give a type hint for a callback

external_samples/component.py
```diff
@@ -1,6 +1,10 @@
 from abc import ABC, abstractmethod
 from enum import Enum
+from collections.abc import Callable, Protocol
```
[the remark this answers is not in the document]
alan412: Oops, when I moved to EmptyCallable, I forgot to remove it.

external_samples/component.py
alan412: oops.   It should have been part of component.  Good catch

external_samples/component.py
alan412: That the last item might be an empty string.   Let me change the comment.

external_samples/rev_touch_sensor.py
alan412: It can't be, because when the system boots up you don't know if the touch sensor is currently pressed or not.   If you start out saying it isn't, then you'll get an immediate callback for pressed even though that isn't accurate.   So you have to start out before your first read with something that knows it hasn't read hardware yet.

Leaving this comment open to see if you agree or not.

lizlooney (approved):

external_samples/rev_touch_sensor.py
lizlooney: OK. That makes sense.

PR #98, opened 2025-03-11
Fixed all warnings except one

lizlooney:
Fixed all warnings except one in App.tsx:  "Could not find a declaration file for module './Blockly'. '/Users/lizlooney/wpilib/1/systemcore-blocks-interface/src/Blockly/index.js' implicitly has an 'any' type."


alan412 (approved):
Tested it.  Thanks for fixing the warnings!

PR #99, opened 2025-03-15
Add sample component for SparkFun LED Stick

lizlooney:


lizlooney:
This is ready for review now.

src/blocks/mrc_call_python_function.ts
```diff
@@ -138,6 +153,12 @@ const CALL_PYTHON_FUNCTION = {
           tooltip = 'Calls the method ' + functionName + '.';
           break;
         }
+        case FunctionKind.INSTANCE_COMPONENT: {
+          const className = this.mrcComponentClassName;
+          const functionName = this.getFieldValue(pythonUtils.FIELD_FUNCTION_NAME);
+          tooltip = 'Calls the function ' + className + '.' + functionName + '.';
```
alan412, line 159: Should the tooltip show the component name instead of the class name?

src/blocks/mrc_class_method_def.ts
```diff
@@ -458,7 +458,8 @@ export const pythonFromBlock = function (
         xfix2 = xfix1;
     }
     if(block.mrcPythonMethodName == '__init__'){
-        branch = generator.defineClassVariables() + branch;
+        branch = generator.INDENT + 'super().__init__()\n' +
```
alan412, line 461: If the init is an opmode, then it needs to pass robot to the super().__init__

alan412 (changes requested):
Why do you create variables for each component used?   It seems like we should just be `self.robot.frontTouch.is_pressed()` so it is clear it is part of the robot instead of `self.frontTouch.is_pressed()`

src/blocks/mrc_class_method_def.ts
alan412: This is assuming that the parent OpMode class should have the line:
`self.robot = robot` in it.

src/blocks/mrc_call_python_function.ts
lizlooney: Yes. Actually I think it should show both, since the name of the component might not overlap with the component class name.

lizlooney:
Thanks for your review comments. PTAL.

external_samples/docs/sparkfun_led_stick.md
lizlooney, line 1: These images are all out-of-date.

src/toolbox/component_samples_category.ts
```diff
@@ -0,0 +1,526 @@
...
+                                type: 'int',
+                            },
+                            {
+                                name: 'color',
+                                type: 'wpilib.Color',
+                            },
+                        ],
+                        tooltip: 'Change the color of an individual LED.',
+                        importModule: '',
+                        componentClassName: 'wpilib.SparkFunLEDStick',
```
lizlooney, line 318: Change wpilib.SparkFunLEDStick to sparkfun.LEDStick.

src/toolbox/component_samples_category.ts
```diff
@@ -0,0 +1,526 @@
...
+                {
+                    kind: 'block',
+                    type: 'mrc_call_python_function',
+                    extraState: {
+                        functionKind: 'instance_component',
+                        returnType: 'bool',
+                        args: [],
+                        tooltip: 'Returns if the touch sensor is pressed or not',
+                        importModule: '',
+                        componentClassName: 'wpilib.RevTouchSensor',
```
lizlooney, line 20: rev.TouchSensor

lizlooney:
Change code generation to use self.robot.<ComponentName>.<function>

src/toolbox/component_samples_category.ts
lizlooney: Done

src/toolbox/component_samples_category.ts
lizlooney: done

external_samples/docs/sparkfun_led_stick.md
lizlooney: done

lizlooney:
OK. Now this is ready for another review.

lizlooney:
> Why do you create variables for each component used? It seems like we should just be `self.robot.frontTouch.is_pressed()` so it is clear it is part of the robot instead of `self.frontTouch.is_pressed()`

Done.

alan412 (approved):
LGTM

PR #100, opened 2025-03-15
Updated generated blocks to fix problem with python class names.

lizlooney:


alan412:
I would *really* like us to figure out a way to get these generated files out of the repo.   Making a change to how things are generated causing a PR with over 400 files changed is yecch.

PR #104, opened 2025-04-07
Use generated json to build toolbox for RobotPy

### LINKED ISSUE #38: Move Generated blocks out of repo
Figure out a way to move the generated blocks out of repo

lizlooney:
Fixes #38


src/blocks/mrc_set_python_variable.ts
```diff
@@ -23,21 +23,27 @@
...
+import { createFieldNonEditableText } from '../fields/FieldNonEditableText';
 import { MRC_STYLE_VARIABLES } from '../themes/styles';
+import * as toolboxItems from '../toolbox/items';
+
 
 // A block to set a python variable.
 
 export const BLOCK_NAME = 'mrc_set_python_variable';
 
+enum VariableKind {
```
alan412, line 39: Why is VariableKind defined in both mrc_get_python_variable.ts and here?   It seems like since they are the same they should be shared.

alan412 (approved):
Great work!  

I had one minor comment about some duplicated code, but if you want to go ahead and merge I am fine.   I suspect we'll have more refactoring ahead of us and that can be done later.

src/blocks/mrc_set_python_variable.ts
lizlooney: I'll put in a TODO so we don't forget.

Issue #113, opened 2025-05-03
Add ability to Hide Python code


alan412:
For small screens, we need a way you can Hide/Show the python code

PR #119, opened 2025-06-14
Change UI around blockly

alan412:
This is a huge PR which changes most of the webpage around Blockly.  It makes no changes to the blockly elements itself.

There is probably a lot here for us to talk about after you get back from vacation, @lizlooney


alan412:
Realized looking through issues that this Fixes #113

lizlooney (approved):

PR #128, opened 2025-06-25
Changes required to fix fast restart

alan412:
Oops, I realized that I had unintentionally broken fast restart by having an enum in a react file.   I pulled it out which required updating a bunch of imports, but the logic is unchanged.

Eventually the type of module shouldn't be a tab type anyway, but that is more of a change than I wanted to do right now.

### DIFF
--- a/src/editor/extended_python_generator.ts
+++ b/src/editor/extended_python_generator.ts
@@ -270,7 +270,7 @@ export class ExtendedPythonGenerator extends PythonGenerator {
         }
       });
       if (exported) {
-        const variableName = variableModel.name;
+        const variableName = variableModel.getName();
         const actualVariableName = super.getVariableName(variableModel.getId());
         const getPythonModuleVariableBlock = {
           'kind': 'block',
--- a/src/editor/extended_python_generator.ts
+++ b/src/editor/extended_python_generator.ts
@@ -270,7 +270,7 @@ export class ExtendedPythonGenerator extends PythonGenerator {
         }
       });
       if (exported) {
-        const variableName = variableModel.getName();
+        const variableName = variableModel.name;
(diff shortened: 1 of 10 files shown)


lizlooney (dismissed):

src/editor/extended_python_generator.ts
```diff
@@ -270,7 +270,7 @@ export class ExtendedPythonGenerator extends PythonGenerator {
         }
       });
       if (exported) {
-        const variableName = variableModel.getName();
+        const variableName = variableModel.name;
```
lizlooney, line 273: Was this change intentional? You changed it to getName() in #124.

src/reportWebVitals.ts
```diff
@@ -1,11 +1,10 @@
 const reportWebVitals = (onPerfEntry?: (metric: any) => void) => {
   if (onPerfEntry && onPerfEntry instanceof Function) {
-    import('web-vitals').then(({ onCLS, onINP, onFCP, onLCP, onTTFB }) => {
-      onCLS(onPerfEntry);
-      onINP(onPerfEntry);
-      onFCP(onPerfEntry);
-      onLCP(onPerfEntry);
-      onTTFB(onPerfEntry);
+    import('web-vitals').then(({ getCLS, getFCP, getLCP, getTTFB }) => {
```
lizlooney, line 3: Same here. This code was part of #124. Do you really want to change it back?

alan412:
I am glad you caught that.   I made both sets of changes based off of what VSCode was telling me were problems, but I think I confused it by switching branches that had different versions in it.   

I looked through things and the changes in #124 were correct.   Let me go add a commit removing those two changes from this one.

alan412:
Glad we squash.  That is an embarrassing set of commits.

PR #131, opened 2025-06-28
Fixed uses of deprecated blockly functions

lizlooney:
Fixed uses of getAllVariables and getVariableUsesById, which are deprecated in v12 and will be deleted in v13.


alan412 (approved):
LGTM

PR #135, opened 2025-06-28
Revert PR 131 because workspace.getVariableMap() doesn't return an ob…

lizlooney:
…ject that has getAllVariables or getVariableUsesById functions.


lizlooney:
I've filed https://github.com/google/blockly/issues/9184 about the deprecation warning suggesting code that doesn't work.

alan412 (approved):
LGTM

PR #140, opened 2025-07-01
Generate blocks for calling component functions

lizlooney:
Added external_samples/spark_mini.py that wraps the wpilib.SparkMini class.

Changed component class names to be consistent with the external_samples python files.

Updated generated json files.

Added function addInstanceComponentBlocks to mrc_call_python_function.ts to generate component function blocks. Updated function getBlocks in blocks_components.ts to use addInstanceComponentBlocks.


src/blocks/mrc_call_python_function.ts
```diff
@@ -612,7 +676,18 @@ export const pythonFromBlock = function(
       const functionName = callPythonFunctionBlock.mrcActualFunctionName
           ? callPythonFunctionBlock.mrcActualFunctionName
           : block.getFieldValue(FIELD_FUNCTION_NAME);
-      code = 'self.' + componentName + '.' + functionName;
+      // Generate the correct code depending on the module type.
+      switch (generator.getModuleType()) {
+        case commonStorage.MODULE_TYPE_PROJECT:
+          code = 'self.';
+          break;
+        case commonStorage.MODULE_TYPE_MECHANISM:
```
alan412, line 684: I don't think this is correct.   The mechanism won't have access to the robot, so I think the mechanism module needs to be with the project one. (which we still need to rename to robot)

src/editor/extended_python_generator.ts
```diff
@@ -248,7 +255,7 @@ export class ExtendedPythonGenerator extends PythonGenerator {
       exportedBlocks.push(callFunctionBlock);
     }
 
-    const allVariables = workspace.getAllVariables();
+    const allVariables = workspace.getVariableMap().getAllVariables();
```
alan412, line 258: Isn't this the same thing you had to revert earlier?

src/toolbox/hardware_components/color_sensor.ts
```diff
@@ -22,7 +22,7 @@
 
 import * as ToolboxItems from '../items';
 
-export const TYPE_NAME = 'ColorRangeSensor';
+export const TYPE_NAME = 'color_range_sensor.ColorRangeSensor';
```
alan412, line 25: Why are the type_names seemingly redundant?   Is it module followed by class?   Do we have modules that have more than one class?

src/editor/editor.ts
```diff
@@ -252,13 +252,13 @@ export class Editor {
...
-    components.push({name: 'clawServo', className: 'rev.Servo'});
-    components.push({name: 'colorSensor', className: 'rev.ColorRangeSensor'});
-    components.push({name: 'ledStick', className: 'sparkfun.LEDStick'});
+    components.push({name: 'frontTouch', className: 'rev_touch_sensor.RevTouchSensor'});
+    components.push({name: 'backTouch', className: 'rev_touch_sensor.RevTouchSensor'});
+    components.push({name: 'leftMotor', className: 'smart_motor.SmartMotor'});
+    components.push({name: 'rightMotor', className: 'smart_motor.SmartMotor'});
+    components.push({name: 'clawServo', className: 'servo.Servo'});
+    components.push({name: 'colorSensor', className: 'color_range_sensor.ColorRangeSensor'});
+    components.push({name: 'ledStick', className: 'sparkfun_led_stick.SparkFunLEDStick'});
```
alan412, line 261: Are these still being used?   If so, where?

alan412 (changes requested):
Are events in the external_samples_data.json file?   

Most of these are minor, but I think there is a bug in the mrc_call_python_function

src/editor/extended_python_generator.ts
lizlooney: getAllVariables is ok to call after getVariableMap(). getVariableUsesById is not ok to call after getVariableMap().

alan412 (approved):
LGTM

PR #141, opened 2025-07-03
Robot now shows ports for a component

alan412:
Fixes the problem you noticed in PR #136


lizlooney (changes requested):
Can you add function comments that explain what getComponentsBlocks and getRobotComponentsBlocks, etc does?

lizlooney (approved):

PR #145, opened 2025-07-04
Populate the dropdown in blocks that call component instance methods

lizlooney:
Moved code that iterates through the components blocks in a holder block from hardware_category.ts to mrc_mechanism_component_holder.ts.

Added code to editor.ts to get the names of components from the robot. Added code to mrc_call_python_function.ts to populate the dropdown in blocks that call component instance methods.


alan412 (approved):
LGTM

PR #146, opened 2025-07-05
Generate the blocks for the +Components toolbox category dynamically.

lizlooney:
Added @classmethod pseudo-constructors (static functions) to python components.
Changed mrc_component to use static functions when instantiating components.
Generate the blocks for the +Components toolbox category dynamically.


src/blocks/mrc_component.ts
```diff
@@ -25,18 +25,43 @@ import { Order } from 'blockly/python';
...
 
+export const FIELD_NAME = 'NAME';
+export const FIELD_TYPE = 'TYPE';
+
+// key is a regex pattern that matches a function argument name, value is type of mrc_port block to use.
+// TODO: Improve these regex pattern.
+// TODO: Should the types here match the PortType values in external_samples/component.py?
+const RECOGNIZED_PORT_ARG_NAME_PATTERNS: { [key: string]: string } = {
+  'can_port': 'SmartCan',
+  'i2c_port': 'SmartIO',
```
alan412, line 45: I2C is its own type of port, it shouldn't be SmartIO

src/blocks/mrc_component.ts
```diff
@@ -25,18 +25,43 @@ import { Order } from 'blockly/python';
...
 export const OUTPUT_NAME = 'mrc_component';
 
+export const FIELD_NAME = 'NAME';
+export const FIELD_TYPE = 'TYPE';
+
+// key is a regex pattern that matches a function argument name, value is type of mrc_port block to use.
+// TODO: Improve these regex pattern.
+// TODO: Should the types here match the PortType values in external_samples/component.py?
+const RECOGNIZED_PORT_ARG_NAME_PATTERNS: { [key: string]: string } = {
+  'can_port': 'SmartCan',
```
alan412, line 44: Why SmartCan?   That isn't what it is called on limelight.

src/blocks/mrc_component.ts
```diff
@@ -25,18 +25,43 @@ import { Order } from 'blockly/python';
...
+  'i2c_port': 'SmartIO',
+  'io_port': 'SmartIO',
+  'smartIO_port': 'SmartIO',
+  'smart_io_port': 'SmartIO',
```
alan412, line 48: Why 3 different ways of identifying SmartIO?

src/blocks/utils/generated/external_samples_data.json
```diff
@@ -698,7 +712,21 @@
...
+            "staticMethods": [
+                {
+                    "args": [
+                        {
+                            "defaultValue": null,
+                            "name": "io_port",
+                            "type": "int"
+                        }
+                    ],
+                    "declaringClassName": "rev_touch_sensor.RevTouchSensor",
```
alan412, line 724: Since this is in JSON that already has className, why does it have to be repeated?

src/blocks/utils/python.ts
```diff
@@ -23,9 +23,9 @@ import { ClassData, PythonData, organizeVarDataByType, VariableGettersAndSetters
 import { robotPyData } from './robotpy_data';
 import { externalSamplesData } from './external_samples_data';
 
-import * as pythonEnum from "../mrc_get_python_enum_value";
-import * as getPythonVariable from "../mrc_get_python_variable";
-import * as setPythonVariable from "../mrc_set_python_variable";
+import * as PythonEnum from "../mrc_get_python_enum_value";
```
alan412, line 26: Why did these get changed from lower case to upper case when in other files you changed them from upper case to lower case?   I don't understand when it should start with lower and when with upper.

alan412 (approved):
The I2C needs to be changed, but after you change that go ahead and merge.

src/blocks/utils/python.ts
lizlooney: I noticed some of yours were upper so I looked it up and Google's ai summary said that upper is preferred for "import * as".

src/blocks/mrc_component.ts
lizlooney: Done.

src/blocks/mrc_component.ts
lizlooney: I updated the values to match the python enum PortType in external_samples/component.py.
That will probably have to be adjusted when mrc_port.ts gets updated.

src/blocks/utils/generated/external_samples_data.json
lizlooney: I'm not sure. Maybe when I wrote the python code that generates the JSON I thought there could be a difference between a function that is inherited from a base class? I'll have to look into that.
I filed https://github.com/alan412/mrc_python_tools/issues/14

src/blocks/mrc_component.ts
lizlooney: I removed that map and just made the code look at the JSON data for component.PortType.

Issue #72, opened 2025-02-25
Update mrc_call_python_function blocks in other modules if a mrc_class_method_def block is changed


lizlooney:
We need to handle changes the the name, parameters, and return type.

PR #157, opened 2025-07-10
Fill the Robot -> Methods category

lizlooney:
Update the Robot -> Methods category on the toolbox when the user is editing an opmode.

Removed code that stored "exportedBlocks".


alan412 (approved):
LGTM.   We should probably log an issue that changing the parameters on a method in the robot doesn't update the opmodes where it is used.

lizlooney:
> LGTM. We should probably log an issue that changing the parameters on a method in the robot doesn't update the opmodes where it is used.

Issue #72 covers that.

PR #159, opened 2025-07-11
When a blocks module is loaded, check mrc_call_python_function blocks.

lizlooney:
When a blocks module is loaded, check mrc_call_python_function blocks.
For component method blocks, check whether the component has been renamed. Update the mrc_call_python_function if necessary. For robot method blocks, check whether the method has been renamed or whether its return value or arguments have changed. Update the mrc_call_python_function if possible. If not possible, add a warning on the block.

Related to #72, but not a complete fix. Changing parameter names is not fixed.

### DIFF
--- a/src/blocks/mrc_call_python_function.ts
+++ b/src/blocks/mrc_call_python_function.ts
@@ -350,7 +350,8 @@ interface CallPythonFunctionMixin extends CallPythonFunctionMixinType {
   mrcActualFunctionName: string,
   mrcClassMethodDefBlockId: string,
   mrcComponentClassName: string,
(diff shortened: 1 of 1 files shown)


lizlooney:
I need to fix a problem where if the user changes the component name (by selecting a different one from the dropdown) on a mrc_call_python_function block, it needs to change the block id that it saves in extra state.

alan412 (approved):
Verified that changing the name works.   If you add a parameter, it could possibly update the one in the opmode to have the additional parameter.  (should be an issue, doesn't need to stop this from merging)

PR #163, opened 2025-07-13
Added python tool that generates json

lizlooney:
Added python tool that generates json

Added support for built-in methods (like print) to mrc_call_python_function.
Use mrc_call_python_function for the print block.

### DIFF
--- a/src/toolbox/test_category.ts
+++ b/src/toolbox/test_category.ts
@@ -14,7 +14,7 @@ export function getCategory(): toolboxItems.Category {
     args: [{
       name: '',
       type: 'str',
-      defaultValue: '',
+      defaultValue: '""',
     }],
   };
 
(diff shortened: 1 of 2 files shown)


src/toolbox/test_category.ts
```diff
@@ -1,11 +1,26 @@
+
 import * as Blockly from 'blockly/core';
-export const getCategory = () => ({
+import * as toolboxItems from './items';
+import { addBuiltInFunctionBlocks } from '../blocks/mrc_call_python_function';
+
+export function getCategory(): toolboxItems.Category {
+  const contents: toolboxItems.ContentsType[] = [];
+
+  const printFunction: FunctionData = {
```
alan412, line 9: FunctionData needs to be imported

alan412 (approved):
I think items.ts needs to have contents as not optional in the Category class.  If contents aren't passed in, then the contents array should be set to [] instead of being allowed to be undefined.   While that isn't part of this PR, this PR causes a typescript error in the test_category file.

That can either be part of this PR, or I can do it in a general cleanup PR that I am planning.

I did find one minor problem, but it doesn't keep me from approving.

src/toolbox/test_category.ts
lizlooney: Done

lizlooney:
> items.ts needs to have contents as not optional in the Category class.

Done

PR #165, opened 2025-07-15
Remove block mrc_print.

lizlooney:
### Before this is merged into main...

If you have modules that use the mrc_print block, you'll need to replace them with the mrc_call_python_function version that looks like this:
Screenshot 2025-07-14 at 8 04 13 PM


alan412 (approved):
LGTM

PR #166, opened 2025-07-15
Clean up methods_category.ts and events_category.ts

lizlooney:
Added function getBaseClassBlocks to mrc_class_method_def.ts 
Modified methods_category to call getBaseClassBlocks for robot, mechanism, and opmode.

Removed src/toolbox/opmode_class_methods.ts.
Removed src/toolbox/robot_class_methods.ts.
Removed src/toolbox/mechanism_class_methods.ts.

Added function createCustomMethodBlock to mrc_class_method_def.ts. 
Modified methods_category to call createCustomMethodBlock instead of hardcoding the mrc_class_method_def block.

Added function createCustomEventBlock to mrc_event.ts. 
Modified event_category.ts to call createCustomEventBlock instead of hardcoding the block.


src/toolbox/methods_category.ts
```diff
@@ -43,6 +40,9 @@ export const getCategory = () => ({
 
 export class MethodsCategory {
   private currentModule: commonStorage.Module | null = null;
+  private robotClassBlocks = getBaseClassBlocks('blocks_base_classes.robot_base.RobotBase');
```
alan412, line 43: Hmmm.   I guess this works now, but if we ever have methods in the base classes that we don't expect to be re-implemented will this cause us a problem?

alan412 (approved):

src/toolbox/methods_category.ts
lizlooney: If that happens we'll decide what to do

PR #168, opened 2025-07-16
Adding Hebrew translation

alan412:
This is from Copilot, so it needs to be verified by a native speaker


lizlooney (commented):
Do we want to put blockly into right-to-left mode?

lizlooney (approved):

PR #170, opened 2025-07-17
Don't generate json for hal.

lizlooney:


alan412 (approved):
LGTM.   I particularly like moving to a list of modules to ignore

PR #172, opened 2025-07-18
Store a version number.

lizlooney:
Store a version number in all modules.

Even if we decide to upgrade when the project is loaded, it's easier to store the version number in every module.


src/storage/common_storage.ts
```diff
@@ -32,6 +32,8 @@ import { createGeneratorContext } from '../editor/generator_context';
 
 // Types, constants, and functions related to modules, regardless of where the modules are stored.
 
+export const CURRENT_VERSION = '0.1';
```
alan412, line 35: I wonder if we should use the blockly version instead of having a version specific for the file format

src/storage/common_storage.ts
```diff
@@ -572,19 +576,26 @@ export function makeModuleContent(
   pythonCode: string,
   blocksContent: string,
   methodsContent: string,
-  componentsContent: string): string {
+  componentsContent: string,
+  optVersionContent?: string): string {
```
alan412, line 580: Not sure we want this optional.   When one is uploaded without it, we are giving it a default value.

PR #175, opened 2025-07-19
Let the user add event handlers to an opmode, for events defined in the robot

lizlooney:
In common_storage.ts, save the events declared in the module.

In mrc_event_handler.ts:
  Refactored hardcoded fields names into constants FIELD_SENDER and FIELD_EVENT_NAME.
  Added otherBlockId to EventHandlerExtraState and mrcOtherBlockId to EventHandlerMixin.
  Added onLoad method that updates event handler blocks if the event definition has been changed.
  Added functions addRobotEventHandlerBlocks and createRobotEventHandlerBlock to create blocks for the toolbox.

In editor.ts, added method getEventsFromRobot.

In hardware_category.ts, added function getRobotEventsBlocks and call it in getHardwareCategory when the current module is an opmode.


lizlooney:
I added another commit here to fix the python code for registering event handlers.

It also fixes a problem where (previously) the hardware ports were collected as the python code was being generated block-by-block, but that didn't work because we need all the ports to generate the call to define_hardware when we generate the code for the __init__ method.

PR #179, opened 2025-07-22
Save JSON only (no python) when saving each module.

lizlooney:
The JSON contains the module type, blocks, components, events, and methods.


alan412 (approved):
I found one place that needed to be changed from py to blocks.   I added it as a commit to your PR.   It now works on my system.   Check to make sure that you don't have any problem with my commit and then merge it.

PR #180, opened 2025-07-25
Upgrade to robotpy 2027.0.0a2.

lizlooney:


alan412 (approved):
Still not thrilled about a massive json file in the repo, but I don't have a better solution right now.

python_tools/generate_json.py
```diff
@@ -91,7 +90,6 @@ def main(argv):
     wpilib.drive,
     wpilib.event,
     wpilib.interfaces,
-    wpilib.shuffleboard,
     wpilib.simulation,
     python_util.getModule('wpilib.sysid'),
     wpimath,
```
auscompgeek, line 95: FWIW this seems to be missing `wpilog`?

PR #183, opened 2025-07-27
Add expansion hub python code.

lizlooney:
Add expansion hub python code.

Improved python tools:
Removed _DICT_FULL_MODULE_NAME_TO_MODULE_NAME. Use <module>.__all__ instead.
Pass the json_generator_robotpy to the constructor for json_generator_external_samples so it can understand the module exports from robotpy.

Renamed SparkMiniComponent to SparkMini.


alan412 (approved):
Looks great, but it took me a while to find the expansion hub stuff running because I expected to see it as a component.    I think that shouldn't block this PR but should be a new one.

PR #185, opened 2025-07-29
Don’t allow 2 event handlers for the same event in the same module

lizlooney:
In common_storage.ts, added function makeUniqueName. Modified makeUploadProjectName to call makeUniqueName.

In editor.ts, added methods getCurrentModuleType and getEventHandlerNames. Removed fields methodsCategory and eventsCategory, but still create all custom toolbox categories in the constructor.

In mrc_event_handler.ts, renamed function getHasEventHandler to getHasAnyEnabledEventHandlers. Added function getEventHandlerNames.

In mrc_event.ts, modified createCustomEventBlock to take the event name as a parameter.

In hardware_category.ts, renamed functions that return a category to be named get...Category instead of get...Blocks. Make RobotEventsCategory a custom category with a flyout callback.

In event_category, removed field currentModule and method setCurrentModule. Make a unique name for creating a new event.

In methods_category, removed field currentModule and method setCurrentModule. Use editor.getCurrentModuleType() instead.

### DIFF
--- a/src/blocks/mrc_event.ts
+++ b/src/blocks/mrc_event.ts
@@ -238,11 +238,11 @@ export const pythonFromBlock = function (
 
 // Functions used for creating blocks for the toolbox.
 
-export function createCustomEventBlock(): toolboxItems.Block {
+export function createCustomEventBlock(name: string): toolboxItems.Block {
   const extraState: EventExtraState = {
     params: [],
   };
   const fields: {[key: string]: any} = {};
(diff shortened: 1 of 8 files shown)


alan412 (changes requested):
This doesn't seem to keep me from having a duplicate handler.
Screenshot 2025-07-29 at 8 20 12 PM

alan412 (approved):
Never mind.   I had forgotten to refresh the browser.   Sigh.

PR #188, opened 2025-08-05
Added components for ExpansionHubServo and ExpansionHubMotor.

lizlooney:
Added components for ExpansionHubServo and ExpansionHubMotor.

Added start and update methods to components.

Added code to run_opmode to print stacktrace if an error occurs.

### DIFF
--- a/external_samples/component.py
+++ b/external_samples/component.py
@@ -71,6 +71,14 @@ class Component(ABC):
     def get_version(self) -> tuple[int, int, int]:
         pass
 
+    @abstractmethod
+    def start(self) -> None:
+        pass
+
+    @abstractmethod
+    def update(self) -> None:
+        pass
+
     # This stops all movement (if any) for the component
     @abstractmethod
     def stop(self) -> None:
--- a/external_samples/expansion_hub_motor.py
+++ b/external_samples/expansion_hub_motor.py
@@ -1,0 +1,126 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This is the expansion_hub_motor module.
+
+This component wraps the expansion_hub.ExpansionHubMotor class, providing
+support for a motor connected to a REV Expansion Hub.
+"""
+
+__author__ = "lizlooney@google.com (Liz Looney)"
+
+from typing import Self
+from component import Component, PortType, InvalidPortException
+import expansion_hub
+import wpimath
+
+# TODO(lizlooney): Update port types.
+
+class ExpansionHubMotor(Component):
+    def __init__(self, ports : list[tuple[PortType, int]]):
+        port_type, hub_number = ports[0]
+        if port_type != PortType.USB_PORT:
+            raise InvalidPortException
+        port_type, motor_number = ports[1]
+        if port_type != PortType.USB_PORT:
+            raise InvalidPortException
+        self.expansion_hub_motor = expansion_hub.ExpansionHubMotor(hub_number, motor_number)
+
+    def get_manufacturer(self) -> str:
+        return "REV Robotics"
+
+    def get_name(self) -> str:
+        return "Expansion Hub Motor"
+
+    def get_part_number(self) -> str:
+        return ""
+
+    def get_url(self) -> str:
+        return ""
+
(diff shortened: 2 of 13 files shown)


external_samples/component.py
```diff
@@ -71,6 +71,14 @@ def get_url(self) -> str:
     def get_version(self) -> tuple[int, int, int]:
         pass
 
+    @abstractmethod
```
alan412, line 74: I think these should be default methods instead of abstract so that components that don't need them don't have to implement them.

external_samples/expansion_hub_motor.py
```diff
@@ -0,0 +1,126 @@
...
+    def getCurrent(self) -> float:
+        return self.expansion_hub_motor.getCurrent()
+
+    def setDistancePerCount(self, perCount: float):
+        self.expansion_hub_motor.setDistancePerCount(perCount)
+
+    def isHubConnected(self) -> bool:
+        return self.expansion_hub_motor.isHubConnected()
+
+    def getEncoder(self) -> float:
```
alan412, line 110: I thought encoder ticks were int instead of float

external_samples/expansion_hub_motor.py
```diff
@@ -0,0 +1,126 @@
...
+        return self.expansion_hub_motor.getEncoder()
+
+    def getEncoderVelocity(self) -> float:
+        return self.expansion_hub_motor.getEncoderVelocity()
+
+    def setReversed(self, reversed: bool):
```
alan412, line 116: Do we need a getReversed as well?

external_samples/expansion_hub_motor.py
```diff
@@ -0,0 +1,126 @@
...
+        self.expansion_hub_motor.setReversed(reversed)
+
+    def resetEncoder(self):
+        self.expansion_hub_motor.resetEncoder()
+
+    def getVelocityPidConstants(self) -> expansion_hub.ExpansionHubPidConstants:
```
alan412, line 122: Do we need a set as well?

external_samples/expansion_hub_motor.py
```diff
@@ -0,0 +1,126 @@
...
+        return self.expansion_hub_motor.getVelocityPidConstants()
+
+    def getPositionPidConstants(self) -> expansion_hub.ExpansionHubPidConstants:
```
alan412, line 125: Do we need a set as well?

external_samples/expansion_hub_servo.py
```diff
@@ -0,0 +1,99 @@
...
+        return cls([(PortType.USB_PORT, hub_number), (PortType.USB_PORT, servo_number)])
+    
+    # Component specific methods
+
+    # Methods from expansion_hub.ExpansionHubServo
+
+    def set(self, value: float):
+        self.expansion_hub_servo.set(value)
+
+    def setAngle(self, degrees: float):
```
alan412, line 86: If this takes degrees, should it be setAngleDegrees?

server_python_scripts/run_opmode.py
```diff
@@ -151,6 +152,7 @@ def run_opmode(opmode_file, duration=None, loop_frequency=50):
         
     except Exception as e:
         print(f"Error running opmode: {e}")
+        traceback.print_exc()
```
alan412, line 155: Great thought to add this!!

alan412 (changes requested):

external_samples/expansion_hub_motor.py
lizlooney: It's float in the class from Thad. See https://github.com/wpilibsuite/systemcore-blocks-interface/blob/485a6a70d054ed7cc26c6a9231ac0afee5852cd4/server_python_scripts/expansion_hub.py#L162

external_samples/expansion_hub_motor.py
lizlooney: Maybe. It will need to be added [here](https://github.com/wpilibsuite/systemcore-blocks-interface/blob/485a6a70d054ed7cc26c6a9231ac0afee5852cd4/server_python_scripts/expansion_hub.py#L176) as well. We should talk to Thad.

external_samples/expansion_hub_motor.py
lizlooney: Maybe. It will need to be added [here](https://github.com/wpilibsuite/systemcore-blocks-interface/blob/485a6a70d054ed7cc26c6a9231ac0afee5852cd4/server_python_scripts/expansion_hub.py#L179) as well. We should talk to Thad.

external_samples/expansion_hub_servo.py
lizlooney: It's just setAngle [here](https://github.com/wpilibsuite/systemcore-blocks-interface/blob/485a6a70d054ed7cc26c6a9231ac0afee5852cd4/server_python_scripts/expansion_hub.py#L216). We can talk about changing it with Thad and others.

server_python_scripts/run_opmode.py
lizlooney: Yeah, I've needed it. haha

alan412 (approved):
I put a question out to thad on slack.  We can fix those items in a later PR after he responds.

PR #190, opened 2025-08-07
Save modules with names like ProjectName.ClassName.json.

lizlooney:
Save modules with names like ProjectName.ClassName.json.
Save modules with multi-line formatted json.

Removed moduleName field from Module (and Robot, Mechanism, and OpMode). Removed userVisibleName from Project.

Renamed ModuleNameComponent.tsx to ClassNameComponent.tsx.

In Menu.tsx, added handleDownload and handleUpload. I didn't add the actual UI element (menu or button, etc) to trigger the handlers.


alan412 (changes requested):
This generates a number of typescript errors.  These are the errors and should be fixed before we merge.

Screenshot 2025-08-07 at 8 05 06 AM

alan412:
Much better, but I still get one error...

Screenshot 2025-08-07 at 9 25 30 PM

PR #195, opened 2025-08-11
Refine Storage interface

lizlooney:
Removed functions related to projects from the Storage interface. Added code to project.ts to handle project stuff.

This will make implementing the server side storage easier.


alan412 (approved):
I am approving this, because I think it takes us in the right direction.   However, I think we will want a listProjects that doesn't read all of the modules.   That way it will return quickly.   I think of it like reading a directory.  You want to return the subdirectory names without parsing all of the files in each subdirectory.

PR #197, opened 2025-08-12
Update the mechanism block in the robot if the mechansim module name has been changed

lizlooney:
Added moduleId to Module and ModuleContent.
Updated mrc_mechanism to find the mechanism by its module id.

### DIFF
--- a/src/blocks/mrc_mechanism.ts
+++ b/src/blocks/mrc_mechanism.ts
@@ -46,6 +46,7 @@ type Parameter = {
 };
 
 type MechanismExtraState = {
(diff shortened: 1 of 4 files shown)


lizlooney:
> This has a bug. If you create a mechanism (say "Arm"), and add it to your robot and then move away from that tab to another and come back it tells you that that mechanism no longer exists.

Good find! There was a bug in the code that recognizes that a module doesn't yet have a module id and it assigns one. But that code might be executed more than once before the mechanism is saved, resulting the the mechanism temporarily being assigned multiple different module ids.

PR #204, opened 2025-08-27
listModulePaths and listProjectNames

lizlooney:
In Storage interface:
Replaced listModules with listModulePaths.
Removed fetchModuleDateModifiedMillis

In projects.ts:
Replace listProjects with listProjectNames.
Added fetchProject.
Changed renameProject, copyProject, deleteProject, and downloadProject functions to take projectName string instead of Project.

In Module type:
Removed dateModifiedMillis, which was never used.

Updated react components Menu, ProjectManageModal, and ProjectNameComponent.


alan412 (approved):
LGTM

PR #206, opened 2025-08-28
Store version number in project.info.json

lizlooney:
Store version number in project.info.json

### DIFF
--- a/src/storage/names.ts
+++ b/src/storage/names.ts
@@ -22,30 +22,80 @@
 import * as storageModule from './module';
 
 
+/**
+ * Paths and file names for Blocks Projects
+ *
+ * Files in a project are stored in a directory whose name is the project name. All files have
+ * the extension '.json' and contain JSON text.
+ *
+ * Project information is stored in a file called 'project.info.json'.
+ *
+ * Modules
+ * Files that contain blocks are called modules. During deployment, a Python file is created for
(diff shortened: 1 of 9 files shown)


src/storage/names.ts
```diff
@@ -22,30 +22,80 @@
...
 export const JSON_FILE_EXTENSION = '.json';
+
+// The extension of a downloaded project is .blocks.
 export const UPLOAD_DOWNLOAD_FILE_EXTENSION = '.blocks';
 
+// The file name of the project info file.
+const PROJECT_INFO_FILE_NAME = 'project.info.json';
+
+// The file name of the project info file.
+const ROBOT_MODULE_FILE_NAME = 'Robot.robot.json';
```
alan412, line 66: Is this necessary for ease of parsing the filename?   If not, I would really prefer just robot.json

alan412 (approved):
I wonder if we should break out the upgrade into its own file.   While it does fairly little now, I can expect this will grow quite a bit in the future.

src/storage/names.ts
lizlooney: Yes, ease of parsing the module file name. All module file names are <ClassName>.<module type>.json.

lizlooney:
> I wonder if we should break out the upgrade into its own file. While it does fairly little now, I can expect this will grow quite a bit in the future.

I've added a TODO to move it.

PR #210, opened 2025-08-29
Share VariableKind enum for get and set python variable blocks

lizlooney:
Share VariableKind enum for get and set python variable blocks.

Remove VariableKind in mrc_set_python_variable.ts.
Use VariableKind from mrc_get_python_variable.ts.


alan412 (approved):
I really like it when we only have an enum in one place. :-)

PR #211, opened 2025-08-29
Use myArray as the variable name for an array

lizlooney:
Use myArray as the variable name for an array.


alan412 (approved):
LGTM

src/blocks/utils/variable.ts
```diff
@@ -76,8 +76,7 @@ export function varNameForType(type: string): string {
   }
   if (type.includes('[')) {
     // The type is an array.
-    // TODO(lizlooney): What's a good name for an array variable?
-    return '';
+    return 'myArray';
```
alan412, line 79: Does this need to be localized?  (or does that create a ton of problems)

lizlooney: I'm not sure. I've added a TODO.

PR #214, opened 2025-08-30
Updated storage interface

lizlooney:
Updated storage interface.

Replaced listFilePaths with list.
Added rename.

### DIFF
--- a/src/storage/project.ts
+++ b/src/storage/project.ts
@@ -32,7 +32,7 @@ import { upgradeProjectIfNecessary } from './upgrade_project';
 export type Project = {
   projectName: string, // For example, WackyWheelerRobot
   robot: storageModule.Robot,
-  mechanisms: storageModule.Mechanism[]
+  mechanisms: storageModule.Mechanism[],
   opModes: storageModule.OpMode[],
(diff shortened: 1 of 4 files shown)


src/storage/project.ts
```diff
@@ -186,13 +182,13 @@ async function renameOrCopyProject(
...
 export async function deleteProject(
     storage: commonStorage.Storage, projectName: string): Promise<void> {
-  const modulePaths: string[] = await storage.listFilePaths(
-      storageNames.makeModulePathRegexPattern(projectName));
+  const projectFileNames: string[] = await storage.list(
+      storageNames.makeProjectDirectoryPath(projectName));
 
-  for (const modulePath of modulePaths) {
-    await storage.deleteFile(modulePath);
+  for (const projectFileName of projectFileNames) {
```
alan412, line 188: Should this logic be above the storage or in the storage?   It might be simpler for the storage to know when you are deleting a directory that you mean to delete everything in it.

lizlooney: Done.

alan412 (approved):
LGTM

PR #220, opened 2025-09-05
Removed i18n string addTabDialog.newItemPlaceholder

lizlooney:
Removed addTabDialog.newItemPlaceholder which was not used in the AddTabDialog.

It was used in the ClassNameCompoonent and ProjectNameComponent but "Add Module" doesn't make sense there.


alan412 (approved):
LGTM

PR #222, opened 2025-09-06
Show button instead of event handler block in toolbox if the event handler has already been used

lizlooney:
Show button instead of event handler block in toolbox if the event handler has already been used.

Modified code that puts event handler blocks in the toolbox so that if an event handler is already on the workspace, a button for that event is put in the toolbox. If the user hits the button, an info message tells the user that the event handler is already on the workspace.

Added class Button to toolbox/items.ts.
Added support for making a category initially expanded.
Cleaned up other classes in toolbox/items.ts.

Modified getHardwareCategory in toolbox/hardware_category.ts so the Hardware or Robot category are initially expanded.

Modified getToolboxJSON in toolbox/toolbox.ts so it prevent a problem with blockly choking on our Category class.


src/blocks/mrc_event_handler.ts
```diff
@@ -405,9 +410,22 @@ function generateRegisterEventHandler(
...
+  // Collect the ids of events for which there is already an event handler.
+  const eventIds: string[] = [];
+  eventHandlerBlocks.forEach(eventHandlerBlock => {
+    eventIds.push(eventHandlerBlock.getEventId());
+  });
   events.forEach(event => {
-    contents.push(createRobotEventHandlerBlock(event));
+    if (eventIds.includes(event.eventId)) {
+      // If there is already an event handler for this event, put a button in the toolbox.
+      const text = '\u00A0\u00A0when\u00A0\u00A0' + SENDER_VALUE_ROBOT + '\u00A0\u00A0' + event.name + '\u00A0\u00A0';
```
alan412, line 423: This needs to go through internationalization both for when and for RTL languages

src/blocks/mrc_event_handler.ts
```diff
@@ -434,9 +452,22 @@ function createRobotEventHandlerBlock(
...
+  // Collect the ids of events for which there is already an event handler.
+  const eventIds: string[] = [];
+  eventHandlerBlocks.forEach(eventHandlerBlock => {
+    eventIds.push(eventHandlerBlock.getEventId());
+  });
   events.forEach(event => {
-    contents.push(createMechanismEventHandlerBlock(mechanismInRobot, event));
+    if (eventIds.includes(event.eventId)) {
+      // If there is already an event handler for this event, put a button in the toolbox.
+      const text = 'when ' + mechanismInRobot.name + ' ' + event.name;
```
alan412, line 465: This needs to go through internationalization both for when and so the RTL languages will work correctly

src/index.css
```diff
@@ -61,4 +61,8 @@ code {
 /* Ensure smooth scrolling for the entire toolbox content */
 .blocklyToolboxContents {
   scroll-behavior: smooth !important;
-}
\ No newline at end of file
+}
+
+.eventHandlerButtonStyle {
+    fill: #995ba5;
```
alan412, line 67: Is it possible (it might not be) to use the style in styles so if a different theme is selected it will still match?

alan412 (changes requested):
I really like this!   A few minor comments.

src/index.css
lizlooney: Done.

PR #236, opened 2025-09-17
Pr hide robot for init

### LINKED ISSUE #228: Hide (robot) on init for opmodes

alan412:
This changes init in opmodes to no longer show robot as a parameter even though it still generates it in the python.

I also moved CURRENT_VERSION and NO_VERSION to the upgrade_project file because that is the file you have to modify when you create a new version.   That way you don't have to touch two different files each time.

Fixes #228

### DIFF
--- a/src/storage/upgrade_project.ts
+++ b/src/storage/upgrade_project.ts
@@ -101,9 +101,9 @@ async function upgradeFrom_001_to_002(
 }
 
 async function upgradeFrom_002_to_003(
-   storage: commonStorage.Storage,
-   projectName: string,
-   projectInfo: storageProject.ProjectInfo): Promise<void> {
+    storage: commonStorage.Storage,
+    projectName: string,
+    projectInfo: storageProject.ProjectInfo): Promise<void> {
   // Opmodes had robot as a parameter to init method
   const projectFileNames: string[] = await storage.list(
     storageNames.makeProjectDirectoryPath(projectName));


alan412:
I apologize for the extra space changes in upgrade_project.ts, I accidentally ran format document on the whole document instead of just my new method.  If you want me to revert those, I can.

src/storage/upgrade_project.ts
```diff
@@ -28,37 +28,44 @@ import * as storageModule from './module';
...
 import * as storageNames from './names';
 import * as storageProject from './project';
+import { ClassMethodDefBlock, Parameter, BLOCK_NAME as MRC_CLASS_METHOD_DEF_BLOCK_NAME } from '../blocks/mrc_class_method_def';
 
+export const NO_VERSION = '0.0.0';
+export const CURRENT_VERSION = '0.0.3';
 
 export async function upgradeProjectIfNecessary(
-    storage: commonStorage.Storage, projectName: string): Promise<void> {
+  storage: commonStorage.Storage, projectName: string): Promise<void> {
```
lizlooney, line 37: Continued lines should be indented 4 spaces.

lizlooney (changes requested):

src/storage/upgrade_project.ts
```diff
@@ -92,3 +99,48 @@ async function upgradeFrom_001_to_002(
...
+      let moduleContentText = await storage.fetchFileContentText(modulePath);
+      const moduleContent = storageModuleContent.parseModuleContentText(moduleContentText);
+      let blocks = moduleContent.getBlocks();
+
+      // Create a temporary workspace to upgrade the blocks
+      const headlessWorkspace = new Blockly.Workspace();
+      try {
+        Blockly.serialization.workspaces.load(blocks, headlessWorkspace);
+
+        // Find and modify init method blocks to remove robot parameter
```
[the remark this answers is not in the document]
alan412: I thought about this, but I didn't want to litter the other files with code we know is temporary.   Do you have a suggestion for a common way for us to mark temporary migration methods so we know they are safe to remove. (and shouldn't be called)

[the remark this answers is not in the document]
lizlooney: Regarding littering the other files... Yes, right now the migration code we are writing is temporary, but eventually (once we share this with alpha testers?), the upgrade code we write won't be temporary.

If you want to call it "migrate" instead of "upgrade", then we can look for that when we want to remove it. We can also add a TODO in upgrade_project.ts to remind us to remove migration code.

I still think lines 124-134 should go in the mrc_class_method_def.ts file because it is using internal knowledge of the mrc_class_method_def block in order to do the migration.

src/storage/upgrade_project.ts
alan412: done

src/storage/upgrade_project.ts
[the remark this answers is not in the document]
alan412: I didn't mean to change verbs from upgrade to migrate.  I kept it as upgrade.   Done now

src/storage/upgrade_project.ts
```diff
@@ -101,9 +101,9 @@ async function upgradeFrom_001_to_002(
 }
 
 async function upgradeFrom_002_to_003(
-  storage: commonStorage.Storage,
-  projectName: string,
-  projectInfo: storageProject.ProjectInfo): Promise<void> {
+   storage: commonStorage.Storage,
```
lizlooney, line 104: You need one more space on these 3 lines.

src/storage/upgrade_project.ts
```diff
@@ -121,17 +121,10 @@ async function upgradeFrom_002_to_003(
...
-          if (block.type === MRC_CLASS_METHOD_DEF_BLOCK_NAME &&
-            block.getFieldValue('NAME') === 'init') {
-            // Remove robot parameter from init method
-            const methodBlock = block as ClassMethodDefBlock;
-            let filteredParams: Parameter[] = methodBlock.mrcParameters.filter(param => param.name !== 'robot');
-            methodBlock.mrcParameters = filteredParams;
-          }
-        }
+        // Method blocks need to be upgraded
+        headlessWorkspace.getBlocksByType(MRC_CLASS_METHOD_DEF_BLOCK_NAME, false).forEach(block => {
```
lizlooney, line 125: Can you move lines 125-127 to a new function (outside of the block class) in mrc_class_method_def.ts?

lizlooney (changes requested):

src/storage/upgrade_project.ts
alan412: done.

src/storage/upgrade_project.ts
alan412: I think it is worse because it makes it more difficult to abstract the upgrade in the future.   I can easily imagine a method that takes a list of block types that need to be upgraded for a certain upgrade and it iterates through all of them calling the method for upgrading.

But I may be missing something and want to understand.

lizlooney: OK. That's a good point.

lizlooney (approved):

PR #239, opened 2025-09-19
Show appropriate warning on block that calls a method on a mechanism's private component

### LINKED ISSUE #229: When you move a component from public to private, the warning that you get “no longer exists” needs to change
Steps to reproduce:
1. Create a mechanism
2. Add a component to the "Components" section in mechanism.

Image

4. Add the mechanism to the robot.
5. In the robot or in an opmode, add a block that uses the mechanism's component.

Image

7. In the mechanism, move the component to the "Private Components" section. Save the mechanism.

Image

8. Reload the browser (This is necessary due to a bug introduced in pr #224.)
9. In the robot (or the opmode), look at the block that uses the mechanism's component.

Image

The message on the block is incorrect. Instead of "...no longer exists", it should say that the component is private in the mechanism.

lizlooney:
Update mrc_call_python_function mrcOnLoad method to check whether a mechanism component is now a private component.

Removed field mrcComponentNames. It only needs to be a local variable in mrcOnLoad.

Added method Editor.getPrivateComponentsFromMechanism.

Fixes #229


alan412 (approved):
LGTM

PR #241, opened 2025-09-19
Cleaned up AddTab Dialog

### LINKED ISSUE #225: Add Tab dialog box needs to be reworked
Make it clearer what to do next

alan412:
This makes the Add Tab Dialog much much simpler and obvious

Fixes #225

### DIFF
--- a/src/i18n/locales/he/translation.json
+++ b/src/i18n/locales/he/translation.json
@@ -36,7 +36,7 @@
   "ABOUT": "אודות",
   "SELECT_HIDDEN": "בחר נסתר",
   "NO_HIDDEN_MECHANISMS": "אין מנגנונים נסתרים",
-  "NO_HIDDEN_OPMODES": "אין מצבי פעולה נסתרים",
+  "NO_HIDDEN_OPMODES": "אין אופמודים נסתרים",
   "CREATE_NEW": "צור חדש",
   "CREATE": "צור",
   "BLOCKS": "בלוקים",


lizlooney:
Can you hide the hidden section if there are no hidden opmodes/mechanism?

Screenshot 2025-09-20 at 9 32 20 PM

src/reactComponents/AddTabDialog.tsx
```diff
@@ -138,34 +112,13 @@ export default function AddTabDialog(props: AddTabDialogProps) {
...
-
-    setNewItemName('');
-  };
-
-  /** Moves an item from available to selected list. */
-  const handleSelectItem = (item: Module): void => {
-    const existingItem = selectedItems.find((i) => i.path === item.path);
-    if (existingItem) {
-      return;
+      props.onOk(newTab);
```
lizlooney, line 120: Can you clear the newItemName so that the next time this dialog is viewed, it doesn't have the previously used name?

lizlooney (changes requested):

alan412:
> This is a big improvement! I just have a couple suggestions.
> 
> Can you hide the hidden section if there are no hidden opmodes/mechanism?
> 
I can but I thought that current UI thought is that it is better to show it as empty rather than have students not understand why it is different on different times they bring it up.

lizlooney:
> > This is a big improvement! I just have a couple suggestions.
> > Can you hide the hidden section if there are no hidden opmodes/mechanism?
> 
> I can but I thought that current UI thought is that it is better to show it as empty rather than have students not understand why it is different on different times they bring it up.

Ok sounds good. What do you think of making the hidden list shorter (less tall) if it is empty?

alan412:
See what you think of it being smaller now if empty

lizlooney (approved):
Looks great!

PR #245, opened 2025-09-23
fixed translation mistakes to hebrew

DanPeled:
renamed OPMODE translation from "action state" (מצב פעולה) to opmode (אופמוד), while isn't a real world it is much more understandable than "action state"

fixed other simmilar translation mistakes


lizlooney:
Thanks so much @DanPeled !

DanPeled:
> Thank you very much for these.   I ran it through copilot and it agreed with your changes.  

Would be concerning if it hadn't, I've been speaking it my entire life 🙂

DanPeled:
Yes, it's understandable

PR #253, opened 2025-09-25
Renamed field is_pressed to _is_pressed in rev_touch_sensor.py

lizlooney:
Renamed field is_pressed to _is_pressed so it doesn't collide with the function is_pressed.


alan412 (approved):
LGTM

PR #257, opened 2025-09-26
Cleaned up mrc_port

### LINKED ISSUE #249: Port for expansion hub servo and motor should be 0, 1, 2, 3, 4, 5 and 0, 1, 2, 3, 4
On an expansion hub, the servo ports are labeled 0, 1, 2, 3, 4, 5, and the motor ports are labeled 0, 1, 2, 3.

Image

The port dropdown for an expansion hub servo shows 1, 2, 3, 4, 5, 6. It should be 0, 1, 2, 3, 4, 5.

Image

The port dropdown for an expansion hub motor shows 1, 2, 3, 4, 5, 6. It should be 0, 1, 2, 3.

Image

lizlooney:
Fixes #249 

In mrc_port.ts.
- Added constants for port types.
- Added constants for field prefixes.
- Added constants for visible port labels.
- Removed type MrcPortType, ports_ from PortMixin, and ports from PortExtraState.
- Renamed PortMixin portType_ to mrcPortType.
- Added mrcPortCount to PortMixin.
- Rewrote loadExtraState to create the appropriate fields for the portType.
- Removed updateShape.
- Simplified the code in pythonFromBlock that collects the ports.
- Generate code for CompoundPort without trailing backslashes.
- Changed Order.ATOMIC to Order.FUNCTION_CALL.
- Updated createPort to specify the field values.

In field_number_dropdown.ts:
Removed defaultVal parameter in createFieldNumberDropdown function.

In external_samples/port.py:
Replace PortType enum value BASE_COMPOUND with a module variable named _base_compound. Removed PortType enum value SERVO_PORT.
Added PortType enum value USB_HUB_PORT.

Removed external_samples/servo.py.


external_samples/port.py
```diff
@@ -19,20 +19,25 @@
 from enum import Enum
 from typing import Self
 
+_base_compound = 256
```
alan412, line 22: I think this should be all caps to show it is supposed to be a constant and shouldn't be changed.  We should probably also import Final from typing and use it here.

alan412 (changes requested):
In general, this looks great and is a huge improvement.   I did have one minor change to the python code.

PR #260, opened 2025-09-27
Make it so you can make methods that return a value

### LINKED ISSUE #251: Allow creating a method that returns something
I can define a new method by dragging the `my_method` block from the Methods category of the toolbox.

Image

However, I can't make it return something.

alan412:
This adds a method that returns a value to the toolbox where you can drag it in and use it to return items.

Fixes #251

### DIFF
--- a/src/blocks/mrc_class_method_def.ts
+++ b/src/blocks/mrc_class_method_def.ts
@@ -38,6 +38,7 @@ import { MUTATOR_BLOCK_NAME, PARAM_CONTAINER_BLOCK_NAME, MethodMutatorArgBlock }
 export const BLOCK_NAME = 'mrc_class_method_def';
 
 export const FIELD_METHOD_NAME = 'NAME';
+export const RETURN_VALUE = 'RETURN';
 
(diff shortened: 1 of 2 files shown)


src/blocks/mrc_class_method_def.ts
```diff
@@ -250,6 +253,21 @@ const CLASS_METHOD_DEF = {
...
+            this.removeInput(RETURN_VALUE);
+        }
+        
+        // Add return input if return type is not 'None'
+        if (this.mrcReturnType && this.mrcReturnType !== 'None') {
+            this.appendValueInput(RETURN_VALUE)
+                .setAlign(Blockly.inputs.Align.RIGHT)
+                .appendField(Blockly.Msg.PROCEDURES_DEFRETURN_RETURN);
+            // Move the return input to be before the statement input
+            this.moveInputBefore(RETURN_VALUE, 'STACK');
```
lizlooney, line 268: Why is the return before the statements?

Image

I think it looks better to have the return after the statements.

Image

PR #261, opened 2025-09-27
Allow code panel to be resized

### LINKED ISSUE #231: python code should be horizontally resizeable

alan412:
This fixes #231.   

It also has some changes to the left menu so that the two would use the same component so they look and act the same.

### DIFF
--- a/src/reactComponents/SiderCollapseTrigger.tsx
+++ b/src/reactComponents/SiderCollapseTrigger.tsx
@@ -1,0 +1,74 @@
+/**
+ * @license
+ * Copyright 2025 Porpoiseful LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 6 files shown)


src/reactComponents/SiderCollapseTrigger.tsx
```diff
@@ -0,0 +1,74 @@
...
+        transition: 'all 0.2s',
+        zIndex: 1,
+      }}
+      onClick={props.onToggle}
+      onMouseEnter={() => setIsHovered(true)}
+      onMouseLeave={() => setIsHovered(false)}
+    >
+      <Antd.Tooltip title={props.collapsed ? t("EXPAND") : t("COLLAPSE")}>
+        {props.collapsed ? 
+          <RightOutlined style={{ fontSize: '12px', color: 'inherit' }} /> :
```
lizlooney, line 68: If the sider is on the left, then > means expand and < means collapse.
But, if the sider is on the right, then < means expand and > means collapse. So, the arrow is pointing the direction that the divider is going to move.

lizlooney (changes requested):
This works well.

I just think the > and < should be swapped if the sider is on the right.

PR #268, opened 2025-09-30
React components i18n

alan412:
This adds i18n support for all react components that were missing it EXCEPT FileManageModal and ProjectManageModal because I plan to refactor those.

### DIFF
--- a/src/reactComponents/ClassNameComponent.tsx
+++ b/src/reactComponents/ClassNameComponent.tsx
@@ -25,6 +25,7 @@ import * as I18Next from 'react-i18next';
 import * as React from 'react';
 import * as commonStorage from '../storage/common_storage';
 import * as storageProject from '../storage/project';
+import * as storageNames from '../storage/names';
 
 /** Props for the ClassNameComponent. */
 interface ClassNameComponentProps {
@@ -66,8 +67,15 @@ export default function ClassNameComponent(props: ClassNameComponentProps): Reac
       return;
     }
 
-    const {ok, error} = storageProject.isClassNameOk(props.project, newClassName, t);
-    if (ok) {
+    let error = '';
+  
+    if (!storageNames.isValidClassName(newClassName)) {
+      error = t('INVALID_CLASS_NAME', { name: newClassName });
+    } else if (storageProject.findModuleByClassName(props.project!, newClassName) != null) {
+      error = t('CLASS_NAME_ALREADY_EXISTS', { name: newClassName });
+    }
+
+    if (!error) {
       clearError();
       props.onAddNewItem();
     } else {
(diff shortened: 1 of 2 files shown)


alan412:
@lizlooney - I am not sure what I think about passing in `t` to the project.ts method.   It seems like perhaps all of the error message should be in what calls it.   Thoughts?

src/reactComponents/Tabs.tsx
```diff
@@ -186,7 +186,7 @@ export function Component(props: TabsProps): React.JSX.Element {
       triggerProjectUpdate();
     } catch (error) {
       console.error('Error renaming module:', error);
-      props.setAlertErrorMessage('Failed to rename module');
+      props.setAlertErrorMessage(t('FAILED_TO_RENAME_MODULE'));
```
lizlooney, line 189: I don't see FAILED_TO_RENAME_MODULE in the i18n files.

src/reactComponents/Tabs.tsx
```diff
@@ -213,7 +213,7 @@ export function Component(props: TabsProps): React.JSX.Element {
 
       if (!originalTab) {
         console.error('Original tab not found for copying:', key);
-        props.setAlertErrorMessage('Original tab not found for copying');
+        props.setAlertErrorMessage(t('MODULE_NOT_FOUND_FOR_COPYING'));
```
lizlooney, line 216: I don't see MODULE_NOT_FOUND_FOR_COPYING in the i18n files.

src/reactComponents/Tabs.tsx
```diff
@@ -223,7 +223,7 @@ export function Component(props: TabsProps): React.JSX.Element {
       triggerProjectUpdate();
     } catch (error) {
       console.error('Error copying module:', error);
-      props.setAlertErrorMessage('Failed to copy module');
+      props.setAlertErrorMessage(t('FAILED_TO_COPY_MODULE'));
```
lizlooney, line 226: I don't see FAILED_TO_COPY_MODULE in the i18n files.

src/reactComponents/Tabs.tsx
```diff
@@ -56,7 +56,7 @@ export interface TabsProps {
 }
 
 /** Default copy suffix for tab names. */
-const COPY_SUFFIX = 'Copy';
+const COPY_SUFFIX = ' (copy)';
```
lizlooney, line 59: This is appended to the module class name in the modal when copying a module. Spaces and parenthesis are not allowed in class names, so this just causes trouble.

Image

alan412:
> > @lizlooney - I am not sure what I think about passing in `t` to the project.ts method. It seems like perhaps all of the error message should be in what calls it. Thoughts?
> 
> I think isClassNameOk should be moved from project.ts to ClassNameComponent.tsx.

Agreed.  I'll move it.

lizlooney (approved):

PR #272, opened 2025-10-04
Changed mechanism, component, and event  blocks to use functions mrcOnMove for handling events from blockly

### LINKED ISSUE #232: When I place a couple of events in the robot and then I drag them to change the order, there are warnings in the console.

### LINKED ISSUE #273: The toolbox should be updated a mechanism, component, or event block is deleted from the holder

lizlooney:
Changed mechanism, component, and event blocks to use functions mrcOnMove for handling events from blockly instead of the change framework.

Renamed mrcValidate to mrcOnModuleCurrent because it is called when the module becomes the current module.

Add warning if a component or mechanism block is moved out of the holder.

Update toolbox if mechanisms, components, or events are moved in to or out of the holder or deleted.

Fixes #232 
Fixes #273


alan412 (approved):
This looks great, but it seems like we should also move mrc_param_container to use this way so that we can completely get rid of ChangeFramework.  Made issue #274

lizlooney:
> This looks great, but it seems like we should also move mrc_param_container to use this way so that we can completely get rid of ChangeFramework. Made issue #274

Agreed. I'm already working on that.

PR #285, opened 2025-10-25
Generate "from <module> import <name>" import statements

### LINKED ISSUE #234: The python generator should be able to generate from x import y

lizlooney:
Generate "from <module> import <name>" import statements.

Also, don't let the user create a module whose name collides with an existing python module.

Fixes #234

### DIFF
--- a/docs/testing_before_pr_checklist.md
+++ b/docs/testing_before_pr_checklist.md
@@ -1,0 +1,22 @@
+For each PR, make sure each of these works
+
+# New items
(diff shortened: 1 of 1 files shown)


alan412 (changes requested):
In general this looks good, but when we put in a component it doesn't use this.   Is that intentional?
Screenshot 2025-10-25 at 8 54 38 AM

alan412 (approved):
LGTM

PR #299, opened 2025-11-07
Fixed mrc_class_method_def blocks for methods that return a value

### LINKED ISSUE #238: Changes to upgrade logic...
This is a place for us to put ideas of upgrade logic, even though we'll wait to do that...

Thought 1 - All of them requiring loading and saving the blocks.   Perhaps that should happen outside of the individual upgrade methods.

Thought 2 - Then perhaps blocks should have a upgradeFrom_xxxx_to_xxx method in them if necessary and the upgrade method just calls those.

Thought 3 - We could even go through all the blocks and if they have a corresponding method call it.

I am pretty convinced of Thought 1 and 2.   Thought 3 is a little inefficient (it would test every block), so right now I am thinking it is not a good idea.

### LINKED ISSUE #296: Block that calls a user defined method won't snap into a socket that expects a boolean or number
Block that calls a user defined method won't snap into a socket that expects a boolean or number.

Until this is fixed, the workaround is to call the function and assign a variable to the return value. Then you can plug the variable into a socket that expects a boolean or number.

lizlooney:
Fixed mrc_class_method_def blocks for methods that return a value.

Also, refactored upgrade functions: I added a new function named upgradeBlocksFiles that takes predicate functions that control what files should be upgraded and functions to do the upgrading. I updated the upgradeFrom_00*_to_00* functions to call upgradeBlocksFiles.

Fixes #296.
Fixes #238.

### DIFF
--- a/src/blocks/mrc_class_method_def.ts
+++ b/src/blocks/mrc_class_method_def.ts
@@ -623,10 +623,6 @@ export function getMethodNamesAlreadyOverriddenInWorkspace(
  * This function should only be called when upgrading old projects.
  */
 export function upgrade_002_to_003(workspace: Blockly.Workspace): void {
-  // Make sure the workspace is headless.
-  if (workspace.rendered) {
-    throw new Error('upgrade_002_to_003 should never be called with a rendered workspace.');
-  }
   workspace.getBlocksByType(BLOCK_NAME).forEach(block => {
     (block as ClassMethodDefBlock).upgrade_002_to_003();
   });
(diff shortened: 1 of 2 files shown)


src/blocks/mrc_class_method_def.ts
```diff
@@ -609,3 +617,31 @@ export function getMethodNamesAlreadyOverriddenInWorkspace(
...
   });
 }
+
+/**
+ * Upgrades the ClassMethodDefBlocks in the given workspace from version 002 to 003.
+ * This function should only be called when upgrading old projects.
+ */
+export function upgrade_002_to_003(workspace: Blockly.Workspace): void {
+  // Make sure the workspace is headless.
+  if (workspace.rendered) {
```
alan412: Should we pull these out and check it in the upgrade project so it doesn't have to be littered throughout the code?

alan412 (commented):
How does this work when there are multiple blocks that need to be upgraded?   Do you just create a method here that calls each one and then pass it in?

src/blocks/mrc_class_method_def.ts
lizlooney: The purpose of checking it here is to make sure this function is not called from code other than the upgrade code. But, since these functions all have names upgrade_*_to_*, I guess it is not necessary.

PR #303, opened 2025-11-10
Pr one workspace per tab

### LINKED ISSUE #283: Put each blockly workspace in its own tab
This is the way that tabs want to work and we just didn't do it this way originally because we were sharing one blockly workspace.

alan412:
I haven't taken it through all the PR checklist because I found that it failed and when I tried it on main it failed there as well.   But since this is pretty simple I am putting it up there for you to start looking at and when we fix the problem in main, it should fix it here as well.

Fixes #283

### DIFF
--- a/src/reactComponents/BlocklyComponent.tsx
+++ b/src/reactComponents/BlocklyComponent.tsx
@@ -228,14 +228,7 @@ export default function BlocklyComponent(props: BlocklyComponentProps): React.JS
     if (workspaceRef.current) {
       if (workspaceRef.current.isVisible() &&
           Blockly.getMainWorkspace().id === workspaceRef.current.id) {
-        // Save scroll position before resize
-        const scrollX = workspaceRef.current.scrollX;
-        const scrollY = workspaceRef.current.scrollY;
-        
         Blockly.svgResize(workspaceRef.current);
-        
-        // Restore scroll position after resize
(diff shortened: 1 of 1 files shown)


src/reactComponents/Tabs.tsx
```diff
@@ -10,7 +10,17 @@
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * WITHOUT WARRANTIES      <Antd.Tabs
```
lizlooney: Uhhh, what happened here?

src/reactComponents/TabContent.tsx
```diff
@@ -0,0 +1,263 @@
...
+    } else {
+      const currentSizePx = typeof codePanelSize === 'string'
+        ? (parseFloat(codePanelSize) / 100) * window.innerWidth
+        : codePanelSize;
+      setCodePanelExpandedSize(currentSizePx);
+      setCodePanelSize(CODE_PANEL_MIN_SIZE);
+      setCodePanelCollapsed(true);
+    }
+
+    setTimeout(() => {
```
lizlooney: There was a comment here. Please put it back.

lizlooney (approved):

PR #312, opened 2025-11-13
Fixed problem where toolbox doesn't show mechanism that was just added to the robot

lizlooney:
Fixed problem where toolbox doesn't show mechanism that was just added to the robot.

Steps to reproduce the problem:
1. Make a new project
2. Add a mechanism "Arm" from the + next to Tabs
3. Click on the Robot tab.
4. Add the mechanism (my_arm) to the Robot.
5. Click on the Teleop tab.
6. The mechanism should be in the toolbox (Robot -> Mechanisms -> my_arm)

Details:
Modified Tabs so it waits for the saveModule to finish before switching tabs.
Modified TabContent so when a module is saved, it updates modulePathToContentText.


alan412 (approved):
LGTM

PR #318, opened 2025-11-17
Make copy/paste work between tabs

### LINKED ISSUE #230: Copy/Paste should work between tabs

### LINKED ISSUE #317: Block that calls a method defined in the same module isn't updated after copy, rename, paste
Steps to reproduce:
1. Create a new project.
2. Add a new method called my_method.
3. Add the "call my_method" block to the init method.
4. Copy the "call my_method" block to the clipboard.
5. Rename the my_method method to your_method. Notice that the "call my_method" block in the init method updates automatically to "call your_method".
6. Paste. The pasted block still says "call my_method". It should say "call your_method".

lizlooney:
Make copy/paste work between tabs.

Fixes #230

Fixes #317


lizlooney:
> In general I really like this, but I found a minor bug. If you copy something that isn't allowed to be copied (say a built-in method or the opmode block) there is no user indication and if you then paste it pastes what was copied before.
> 
> I think it should tell you that you can't copy that and also clear the clipboard.

Is that minor bug introduced in this PR? Or, is that minor bug present in the main branch?

PR #327, opened 2025-11-22
Update python when blocks change

### LINKED ISSUE #315: Python code isn't updated when I change the blocks

lizlooney:
Update python when blocks change.

Fixes #315


alan412 (approved):
I am not entirely sure I understand why this fixes it, but it does.   There is still much I need to learn about react.....   anways, LGTM

PR #329, opened 2025-11-22
Pr backend serve

alan412:
This makes a backend that will serve the frontend and also do deployment.

To use it:
1. npm run build (creates a dist directory)
2. source backend/venv/bin/activate
3. python backend/main.py -p xxxx (port you want)

If you press deploy and you are running it from the npm run dev, then it will work like it has been where deploy downloads a zip file.   If you press deploy and you are running it from the python backend script then it will unzip the zip file in a deploy directory under the directory the backend runs from.

### DIFF
--- a/src/reactComponents/Menu.tsx
+++ b/src/reactComponents/Menu.tsx
@@ -294,7 +294,7 @@ export function Component(props: MenuProps): React.JSX.Element {
     }
 
     try {
-      const blobUrl = await createPythonFiles.producePythonProjectBlob(props.project, props.storage);
+      const blobUrl = await createPythonFiles.producePythonProjectBlob(props.currentProject, props.storage);
       
       // Check if the backend server is available
       const serverAvailable = await serverSideStorage.isServerAvailable();
@@ -306,7 +306,7 @@ export function Component(props: MenuProps): React.JSX.Element {
         const blob = await response.blob();
         
         const formData = new FormData();
-        formData.append('file', blob, `${props.project.projectName}.zip`);
+        formData.append('file', blob, `${props.currentProject.projectName}.zip`);
         
         const deployResponse = await fetch('/deploy', {
           method: 'POST',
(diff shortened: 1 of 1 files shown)


backend/main.py
```diff
@@ -224,25 +229,153 @@ def get(self):
...
+        'status': 'ok',
+        'server': 'python-backend',
+        'version': '1.0'
     })
 
+# Handle the base path for the frontend
+@app.route('/blocks/', defaults={'path': ''})
+@app.route('/blocks/<path:path>')
+@app.route('/blocks//<path:path>')  # Handle double slash
+def serve_frontend(path):
```
lizlooney: Please add python 3 type hints.

backend/main.py
```diff
@@ -224,25 +229,153 @@ def get(self):
         
         return {'files': sorted(list(children))}
 
+class DeployResource(Resource):
```
lizlooney: I think you should break this file into separate files. Maybe put all the Storage... classes in one file and put DeployResource in another file.

backend/main.py
```diff
@@ -224,25 +229,153 @@ def get(self):
...
 api.add_resource(StorageEntryResource, '/entries/<path:entry_key>')
 api.add_resource(StorageFileRenameResource, '/storage/rename')
 api.add_resource(StorageRootResource, '/storage/')
 api.add_resource(StorageResource, '/storage/<path:path>')
+api.add_resource(DeployResource, '/deploy')
 
-@app.route('/')
-def index():
+# API health check endpoint to distinguish from static file serving
+@app.route('/api/status')
```
lizlooney: Maybe call this statusz.
Look up z pages or "statusz" if you are not familiar. I found an example here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/4827-component-statusz/README.md#sample-response-1

src/storage/server_side_storage.ts
```diff
@@ -31,13 +31,18 @@ export async function isServerAvailable(): Promise<boolean> {
...
     
-    // Race between the fetch and timeout
+    // Check the specific API status endpoint to distinguish backend from static file server
+    // Use absolute path without base URL since /statusz is a backend endpoint
     const response = await Promise.race([
-      fetch(`${API_BASE_URL}/`),
+      fetch('/statusz'),
       timeoutPromise
-    ]);
+    ]);    if (!response.ok) {
```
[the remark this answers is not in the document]
alan412: oops.  Accidentally got deleted

PR #331, opened 2025-11-22
Change the name of a component in the toolbox when the name field in a Component block is changed

### LINKED ISSUE #289: Changing the name of a component in a robot doesn't change in toolbox until saved...
Steps to reproduce:
 1. Put motor in components on robot
 2. Change name to shooter
 3. See that it isn't changed in toolbox (but is where it is used other places)

Image

lizlooney:
Change the name of a component in the toolbox when the name field in a Component block is changed.

Fixes #289


alan412 (approved):
LGTM.  I found a couple bugs in testing but they aren't related to this PR so I am filing them as issues.

PR #334, opened 2025-11-24
Fix bug with deleting the current project

### LINKED ISSUE #316: Deleting the current project makes the whole window go white
Steps to reproduce:
1. Create a new project named "One".
2. Create another new project named "Two". Project Two is now the current project.
3. Click Manage | Projects...
4. Click the delete button (trash can) next to project Two.
5. Click "Delete" in the confirmation dialog.

The whole window goes white.

The console shows `Uncaught TypeError: Cannot read properties of undefined (reading 'getBlocks')` occurring on [this line](https://github.com/wpilibsuite/systemcore-blocks-interface/blob/3fe6f5c9ed3c1f032f7a12004883a06d35850fde/src/editor/editor.ts#L249).
```
Blockly.serialization.workspaces.load(moduleContent.getBlocks(), this.blocklyWorkspace);
```

lizlooney:
Modified handleDeleteProject to select the next project after deleteProject has finished.

Fixes #316

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -499,8 +499,8 @@ const AppContent: React.FC<AppContentProps> = ({ project, setProject }): React.J
                 storage={storage}
                 setAlertErrorMessage={setAlertErrorMessage}
                 gotoTab={setActiveTab}
-                project={project}
-                setProject={setProject}
+                currentProject={project}
+                setCurrentProject={setProject}
                 onProjectChanged={onProjectChanged}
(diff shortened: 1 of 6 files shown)


lizlooney:
I created this as a draft PR because I'm not sure that we should select another project and exit the dialog when the user deletes a project. What if they delete a project that isn't even the current project? What if they want to delete several project?
Let's talk about this on Monday.

PR #358, opened 2025-12-19
Removed Component base class and external_samples

### LINKED ISSUE #356: Remove Component base class
- Remove opmode_start, opmode_periodic, and opmode_end methods from external_samples python classes.
- Instead, do stuff in their constructors (instead of opmode_start), create background thread (instead of opmode_periodic), and are closable (instead of opmode_end).
- Put some kind of marker in the doc strings that indicates that the class represents a hardware device (what we've been calling a "component") that can be added to a robot.
- Constructors should take arguments that are not just numbers. If there were classes that identified that a numeric value is actually a CAN port or a USB port, etc (similar to wpimath.units.milliseconds, but for types of ports on the Systemcore or Motioncore or ExpansionHub), then blocks would know how to generate code to instantiate the hardware device.

lizlooney:
Removed Component base class and external_samples.

Removed opmode_start, opmode_periodic, and opmode_end methods.

Modified python tools to produce information about components, including whether a class is a component, the component port name (which will show next to the input socket on the mrc_component block), and the component port type (a single string consisting of one or more port types separated by two underscores).

Fixes #356

### DIFF
--- a/src/blocks/mrc_port.ts
+++ b/src/blocks/mrc_port.ts
@@ -46,15 +46,18 @@ const VISIBLE_PORT_LABEL_SERVO = 'servo';
 
 export type PortBlock = Blockly.Block & PortMixin;
 interface PortMixin extends PortMixinType {
-  mrcPortTypeArray: storageModuleContent.PortType[],
+  mrcPortTypes: storageModuleContent.PortType[],
 }
 type PortMixinType = typeof PORT;
 
 type PortExtraState = {
+  // TODO: Remove portType in the near future.
   // A single string consisting of one or more port types. Multiple port types are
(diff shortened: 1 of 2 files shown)


src/modules/mechanism_start.json
```diff
@@ -20,22 +20,6 @@
                     "NAME": "init"
                 }
             },
-            {
```
alan412: Why are we getting rid of the periodic method for mechanisms?   I don't think we want students to be creating a background thread and they need someway of firing events

alan412 (commented):
Let's talk about this on Monday and how we want to proceed

PR #363, opened 2025-12-24
Pr parameter fixes

### LINKED ISSUE #293: Get parameter needs dropdown
For get parameter, allow changing to another parameter with a dropdown

### LINKED ISSUE #295: Deleting used parameter
If you delete a parameter that is used, you need to put an error on that parameter block

alan412:
Fixes #293 (Adding a drop down to parameters)
Fixes #295 (Deleting a used parameter should flag where that parameter is being used as an error)

### DIFF
--- a/src/blocks/mrc_get_parameter.ts
+++ b/src/blocks/mrc_get_parameter.ts
@@ -89,15 +89,15 @@ const GET_PARAMETER_BLOCK = {
     this.setOutput(true, this.mrcParameterType);
   },
   getParameterOptions: function(this: GetParameterBlock): [string, string][] {
-    const legalParameterNames: string[] = [];
+    const existingParameterNames: string[] = [];
 
     const rootBlock: Blockly.Block | null = this.getRootBlock();
     if (rootBlock && rootBlock.type === MRC_CLASS_METHOD_DEF) {
       const classMethodDefBlock = rootBlock as ClassMethodDefBlock;
-      legalParameterNames.push(...classMethodDefBlock.mrcGetParameterNames());
+      existingParameterNames.push(...classMethodDefBlock.mrcGetParameterNames());
     } else if (rootBlock && rootBlock.type === MRC_EVENT_HANDLER) {
       const eventHandlerBlock = rootBlock as EventHandlerBlock;
-      legalParameterNames.push(...eventHandlerBlock.mrcGetParameterNames());
+      existingParameterNames.push(...eventHandlerBlock.mrcGetParameterNames());
     }
 
     // Get the field to check its value
@@ -105,15 +105,15 @@ const GET_PARAMETER_BLOCK = {
     const currentValue = field?.getValue();
     
     // Always include the current field value if it exists and isn't already in the list
-    if (currentValue && currentValue !== '' && !legalParameterNames.includes(currentValue)) {
-      legalParameterNames.unshift(currentValue);
+    if (currentValue && !existingParameterNames.includes(currentValue)) {
+      existingParameterNames.unshift(currentValue);
     }
 
-    if (legalParameterNames.length === 0) {
+    if (existingParameterNames.length === 0) {
       return [[Blockly.Msg.NO_PARAMETERS, '']];
     }
 
-    return legalParameterNames.map(name => [name, name]);
+    return existingParameterNames.map(name => [name, name]);
   },
   validateParameterSelection: function(this: GetParameterBlock, newValue: string): string {
     // Clear any previous warnings


src/blocks/mrc_event_handler.ts
```diff
@@ -134,6 +135,9 @@ const EVENT_HANDLER = {
       });
     });
     this.mrcUpdateParams();
+    
+    // Update all mrc_get_parameter blocks to recheck validity
+    this.mrcCheckParameterBlocks();
```
lizlooney: Is this necessary? Since mrcOnLoad calls checkEvent, which calls mrcCheckParameterBlocks, I don't think this is necessary.
Also, I'm concerned that the parameter blocks might not be fully formed at the time that this block's loadExtraState is called.

src/blocks/mrc_get_parameter.ts
```diff
@@ -70,11 +69,73 @@ const GET_PARAMETER_BLOCK = {
     this.mrcHasWarning = false;
 
     this.setStyle(MRC_STYLE_VARIABLES);
+    
+    // Create a custom dropdown that accepts any value and displays it correctly
```
lizlooney: I don't understand what this mean by "accepts any value". Can you make the comment more clear? How is the value "accepted"? Does the user type in the value?

lizlooney: Also what does "displays it correctly" mean? Why wouldn't it be displayed correctly?

src/blocks/mrc_get_parameter.ts
```diff
@@ -70,11 +69,73 @@ const GET_PARAMETER_BLOCK = {
...
+      const eventHandlerBlock = rootBlock as EventHandlerBlock;
+      legalParameterNames.push(...eventHandlerBlock.mrcGetParameterNames());
+    }
+
+    // Get the field to check its value
+    const field = this.getField(FIELD_PARAMETER_NAME) as Blockly.FieldDropdown | null;
+    const currentValue = field?.getValue();
+    
+    // Always include the current field value if it exists and isn't already in the list
+    if (currentValue && currentValue !== '' && !legalParameterNames.includes(currentValue)) {
```
lizlooney: I think `&& currentValue !== ''` is redundant in typescript/javascript.

src/blocks/mrc_get_parameter.ts
```diff
@@ -70,11 +69,73 @@ const GET_PARAMETER_BLOCK = {
...
+    dropdown.setValidator(this.validateParameterSelection.bind(this));
+    
     this.appendDummyInput()
         .appendField(Blockly.Msg.PARAMETER)
-        .appendField(createFieldNonEditableText(''), FIELD_PARAMETER_NAME);
+        .appendField(dropdown, FIELD_PARAMETER_NAME);
     this.setOutput(true, this.mrcParameterType);
   },
+  getParameterOptions: function(this: GetParameterBlock): [string, string][] {
+    const legalParameterNames: string[] = [];
```
lizlooney: Is "legal" the right word here? Does this array contain "legal" names, or the method/event's parameter names?
If you can think of a better name, great. Otherwise, please add a comment explaining what "legal" means in this context.

src/blocks/mrc_get_parameter.ts
```diff
@@ -70,11 +69,73 @@ const GET_PARAMETER_BLOCK = {
     this.mrcHasWarning = false;
 
     this.setStyle(MRC_STYLE_VARIABLES);
+    
+    // Create a custom dropdown that accepts any value and displays it correctly
+    class CustomParameterDropdown extends Blockly.FieldDropdown {
+      override doClassValidation_(newValue?: string): string | null {
```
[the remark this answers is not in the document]
lizlooney: This class is pretty much exactly the same as CustomStepDropdown in your other PR. You should definitely put the class in it's own file.

src/blocks/mrc_event_handler.ts
alan412: You are right.  It isn't necessary (and could be problematic).  I removed it

src/blocks/mrc_event_handler.ts
```diff
@@ -328,6 +336,23 @@ const EVENT_HANDLER = {
     });
     return parameterNames;
   },
+
+  /**
+   * Checks all mrc_get_parameter blocks within this event handler to revalidate
+   * that their parameter names are still valid.
+   */
+  mrcCheckParameterBlocks: function(this: EventHandlerBlock): void {
```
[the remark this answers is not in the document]
alan412: done

PR #368, opened 2025-12-27
Added a few more components, including some that have non-port arguments

### LINKED ISSUE #366: Support components whose constructor does not take a port, but takes other arguments
wpilib.OnboardIMU should be a component. However, its constructor does not take a port. It takes a wpilib.OnboardIMU.MountOrientation.

Our component code currently assumes that a component constructor takes one or more ports. We need to handle this case where the constructor parameter is a wpilib.OnboardIMU.MountOrientation.

### LINKED ISSUE #367: Support components whose constructors take additional arguments
[AnalogEncoder](https://robotpy.readthedocs.io/projects/robotpy/en/stable/wpilib/AnalogEncoder.html#wpilib.AnalogEncoder) has multiple constructors and the one we probably want to use is this
```
__init__(self: wpilib._wpilib.AnalogEncoder, channel: typing.SupportsInt, fullRange: typing.SupportsFloat, expectedZero: typing.SupportsFloat) -> None
```

lizlooney:
Generate JSON componentArgs for component constructors, instead of componentPortName and componentPortType. Fixed python code that recognizes enums.

Moved PortType and related functions from module_content.ts to python_json_types.ts

Increment the current version to 0.0.10. Upgrade old projects so that Components have args instead of ports.

Changed blocks and the python generator where it used to assume that all components have a port argument and no other arguments.

Fixed the code that generates the robot when it creates a mechanism, passing one tuple of args for each component in the mechanism.
    
Fixed the code that generates a mechanism's __init__ and define_hardware methods. The __init__ and define_hardware methods take one parameter (which is is a tuple) for each component in the mechanism. The define_hardware method uses the * operator to unpack the elements of the tuple and pass each element as a separate positional argument to the component constructor.

Fixes #366
Fixes #367

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -29,6 +29,7 @@ import SiderCollapseTrigger from './reactComponents/SiderCollapseTrigger';
 import ToolboxSettingsModal from './reactComponents/ToolboxSettings';
 import * as Tabs from './reactComponents/Tabs';
 import { TabType } from './types/TabType';
+import { AutosaveProvider } from './reactComponents/AutosaveManager';
 
 import { extendedPythonGenerator } from './editor/extended_python_generator';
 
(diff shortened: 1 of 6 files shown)


alan412 (approved):
How did you test all of these changes?   The code looks reasonable to me, and if you feel like it is tested well I am good with your merging it.

PR #373, opened 2026-01-02
Pr upload download

### LINKED ISSUE #362: Add download / upload of entire blocks project

alan412:
Fixes #362 

This uses the already existing download/upload but adds a UI around it including name conflict resolution on upload.

Note that the automatic changing to the new project exposes a bug (fixed in #372) but that is not a bug in this PR.

After downloading and uploading the same, I went through the pr checklist and it all passes on my machine.
(I substituted the upload for the making a new project)

### DIFF
--- a/src/storage/project.ts
+++ b/src/storage/project.ts
@@ -382,16 +382,6 @@ export async function downloadProject(
   return URL.createObjectURL(content);
 }
 
-/**
- * Make a unique project name for an uploaded project.
- */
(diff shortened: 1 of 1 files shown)


src/reactComponents/Menu.tsx
lizlooney: Remove this old code.

lizlooney: Remove the TODO comment.

PR #378, opened 2026-02-05
Tell user which block was executing when a crash occurs

### LINKED ISSUE #375: Tell user which block was executing when a crash occurs
In FTC SDK, if you run a Blocks OpMode and it crashes, we report which block was executing (or which block had just finished executing) when the crash occurred. We need to do this for Systemcore Blocks.

Notes:
Probably the way to do this is to generate some python code at the beginning and end of each block that saves and clears some information. Then when the crash is handled, the handler will read that information.
If possible, we should only generate the extra code when we are deploying, not when saving the modules and definitely not for the python code that is shown next to the blocks editor.

lizlooney:
Generate python code for error handling, but only for deployed python code.

Fixes #375

### DIFF
--- a/server_python_scripts/run_opmode.py
+++ b/server_python_scripts/run_opmode.py
@@ -152,7 +152,6 @@ def run_opmode(opmode_file, duration=None, loop_frequency=50):
         
     except Exception as e:
         print(f"\nError running opmode: {e}")
-        #traceback.print_exc()
         BlockExecution.handleFatalError(e)
         sys.exit(1)
 


server_python_scripts/run_opmode.py
```diff
@@ -151,8 +151,9 @@ def run_opmode(opmode_file, duration=None, loop_frequency=50):
         print(f"  Average frequency: {actual_frequency:.1f} Hz")
         
     except Exception as e:
-        print(f"Error running opmode: {e}")
-        traceback.print_exc()
+        print(f"\nError running opmode: {e}")
+        #traceback.print_exc()
```
alan412: Why is this commented code left in?

PR #387, opened 2026-03-23
Simplify menu

alan412:
This moves the project download to the manage projects, gets rid of save (since we have autosave), and moves deploy to the top level (and gives it an icon)


lizlooney:
I don't think we should get rid of Save.

What if a kid is coding and their parent comes to pick them up. The parent wants the kid to hurry up, but the kid has to wait for auto-save to trigger!

lizlooney (changes requested):

lizlooney (approved):

PR #393, opened 2026-04-12
Fixed python tools to handle unqualified type aliases in function signatures

lizlooney:
I found this problem when I was working on separating the package from the class name in function blocks.

Examples:
- The module wpimath.units defines a type alias called `degrees`. The qualified name is `wpimath.units.degrees`
  - The function wpimath.units.degreesToRotations takes`degrees`.
  - The function wpimath.units.rotationsToDegrees returns `degrees`.


alan412 (approved):
LGTM

PR #397, opened 2026-04-20
Show simple class names on blocks

lizlooney:
This feature is on by default, but the user can change it from the Settings menu.

I ran through the checklist.

I also tested it with a project that uses all the types of blocks that show class names:
With full names:
Screenshot 2026-04-19 at 10 06 42 PM
With simple names:
Screenshot 2026-04-19 at 10 06 57 PM


alan412 (approved):
LGTM.    At first, I didn't like the check mark in front, but that is what Apple does.  I wonder if there is some way we could show an "empty" icon so the line doesn't move in or out.  But I think that is a minor nit and something we could deal with later.

PR #398, opened 2026-04-21
Fix up some paths in the python tools

lizlooney:
Fix up some paths in the python tools.


alan412 (approved):
LGTM.  Sorry I missed these with the moving things around.

PR #399, opened 2026-04-23
Use camelCase instead of snake_case

lizlooney:
Use camelCase instead of snake_case for:
- robot variables, methods, and arguments
- component constructor arguments
- mechanism variables, methods, and arguments
- default mechanism name
- default name when creating a new event
- default name when creating a new method

I ran through the checklist and I tested a project on my systemcore.


alan412 (approved):
LGTM

PR #421, opened 2026-05-07
Fix Close other behavior

### LINKED ISSUE #413: Close other tabs should only close the ones on the same row
Right now selecting close other tabs closes all tabs other than the current one (on both rows) which doesn't seem correct

alan412:
This fixes it so that selecting close others on a tab only affects the ones in the same row.

Fixes #413

### DIFF
--- a/frontend/reactComponents/Tabs.tsx
+++ b/frontend/reactComponents/Tabs.tsx
@@ -330,12 +330,13 @@ export const Component = React.forwardRef<TabsRef, TabsProps>((props, ref): Reac
     const isRow2 = currentTab.type === TabType.OPMODE;
     const newTabs = props.tabList.filter((tab) => {
       if (tab.key === currentTabKey) return true;
+      if (tab.type === TabType.ROBOT) return true; // Always keep ROBOT tabs
       if (isRow2) {
-        // Keep all row 1 tabs (ROBOT and MECHANISM)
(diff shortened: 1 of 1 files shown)


frontend/reactComponents/Tabs.tsx
```diff
@@ -326,15 +323,34 @@ export const Component = React.forwardRef<TabsRef, TabsProps>((props, ref): Reac
...
   const handleCloseOtherTabs = (currentTabKey: string): void => {
-    const newTabs = props.tabList.filter(
-      (tab) => tab.key === currentTabKey || tab.type === TabType.ROBOT
-    );
+    const currentTab = props.tabList.find((tab) => tab.key === currentTabKey);
+    if (!currentTab) return;
+    const isRow2 = currentTab.type === TabType.OPMODE;
+    const newTabs = props.tabList.filter((tab) => {
+      if (tab.key === currentTabKey) return true;
+      if (isRow2) {
```
lizlooney, line 334: What do you think of:
```
      if (tab.type === TabType.ROBOT) {
        // The robot tab is not closeable.
        return true;
      }
      if (isRow2) {
        // Keep all MECHANISM tabs.
        return tab.type === TabType.MECHANISM;
      } else {
        // Keep all OPMODE tabs.
        return tab.type === TabType.OPMODE;
      }
```

lizlooney (changes requested):

lizlooney (approved):

PR #426, opened 2026-05-15
Add isCommon to FunctionData and VarData

lizlooney:
Add isCommon to FunctionData and VarData

Set isCommon on some methods in the hardcoded components.


alan412 (approved):
Not sure I agree with which methods are marked as common, but since it is only in the generated I am assuming this is just for testing.  (ie, as we do it for "real" we'll re-generate that data.)

PR #429, opened 2026-05-29
Make wpiutil.Color block look awesome!

lizlooney:
Here's what it looks like:

Screenshot 2026-05-27 at 11 43 48 PM

Here's what it looks like as a shadow block:

Screenshot 2026-05-28 at 7 40 42 PM


alan412 (approved):
This looks and works great!

PR #440, opened 2026-06-22
Versioning changes

alan412:
This changes so that all versioning now comes off of what is in package.json

This also removes all old versioning code (but keeps the helpers)


frontend/storage/upgrade_project.ts
```diff
@@ -27,121 +27,43 @@ import * as storageModule from './module';
...
-        await upgradeFrom_0014_to_0015(storage, projectName, projectInfo);
-    }
-    await storageProject.saveProjectInfo(storage, projectName, projectInfo);
-  }
+  projectInfo.version = CURRENT_VERSION;
+  await storageProject.saveProjectInfo(storage, projectName, projectInfo);
 }
 
-async function upgradeBlocksFiles(
+export async function upgradeBlocksFiles(
```
lizlooney: Lots of functions now have `export` but they aren't used outside of this file.

frontend/storage/upgrade_project.ts
```diff
@@ -189,244 +111,36 @@ async function upgradeBlocksFiles(
   }
 }
 
-/**
- * Predicate function that can be passed to upgradeBlocksFiles indicating that all modules should be
- * affected.
- */
-function anyModuleType(_moduleType: storageModule.ModuleType): boolean {
+/** Predicate: all module types are affected. */
```
lizlooney, line 112: The comments are more concise now. That's fine, but I'm curious whether that was your change or Claude's.

lizlooney (changes requested):

frontend/storage/upgrade_project.ts
alan412: They have to be or they will generate TS warnings for being unused.   I didn't want to get rid of the helper methods, and I didn't want warnings.   Are there better ways to silence the warnings?

frontend/storage/upgrade_project.ts
alan412: Claude suggested it, and when I reviewed it it seemed fine to me.

frontend/storage/upgrade_project.ts
alan412: Doing some research, I found a better way.

lizlooney (approved):

PR #449, opened 2026-06-25
Fix packaging

alan412:
This removes the last instance of building the installer from source and instead getting the pre-built one from pypi.

It also adds making the directory since rsync will only make the last directory in a tree.


packaging/buildVenv.sh
```diff
@@ -7,8 +7,6 @@ WHEELS_DIR="$SCRIPT_DIR/wheels_venv"
 PYTHON_VER="3.13"
 SITE_PACKAGES="$VENV_DIR/lib/python${PYTHON_VER}/site-packages"
 VENV_BIN="$VENV_DIR/bin"
-INSTALLER_SRC="$HOME/Projects/mrc-work/robotpy-installer"
```
lizlooney, line 10: Do I have to have this? Or is this for github actions?

lizlooney (changes requested):
Can you change the description (the first comment) of the PR
https://github.com/wpilibsuite/systemcore-blocks-interface/pull/449#issue-4739407763
to explain what this PR is changing? (Imagine us reading it in a year and trying to remember what it was about.)

packaging/buildVenv.sh
alan412: The buildVenv.sh is for when you do npm run package.    This creates the ipk file.

lizlooney (approved):

PR #456, opened 2026-06-25
Use server storage

### LINKED ISSUE #455: Blocks uses local (browser) storage even when it is connected to a backend
This is bad because it means the code isn't on the robot.

alan412:
This PR will require a new release.  It modifies the version of blocks to 0.1.1

It turns out there was a bug where we always used the browser storage even if the backend was available.   When that bug was fixed, it exposed a bug where the db wasn't created if the backend was started with gunicorn (as we do on systemcore).   

That bug was fixed.   This has been tested on Systemcore with the package made.

Fixes #455

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -133,8 +133,10 @@ def serve_static(path: str) -> Union[Response, Tuple[Response, int]]:
         except:
             return jsonify({'error': 'File not found'}), 404
 
+# This is only run for testing.   When we use gunicorn, this ends up not being
+# run.   So be careful of what code is placed here.
 if __name__ == '__main__':
-    parser = argparse.ArgumentParser(description='Run the Storage API backend server')
+    parser = argparse.ArgumentParser(description='Run the Blocks backend server')
     parser.add_argument('-p', '--port', type=int, default=5001,
                         help='Port to run the server on (default: 5001)')
     parser.add_argument('-l', '--log-level', type=str, default='INFO',
--- a/backend/main.py
+++ b/backend/main.py
@@ -42,6 +42,10 @@ app.config['SQLALCHEMY_DATABASE_URI'] = f'sqlite:///{os.path.join(basedir, "proj
 db.init_app(app)
 app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
 
+# Initialize DB tables at startup regardless of how the server is launched (e.g. gunicorn)
+with app.app_context():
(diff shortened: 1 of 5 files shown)


backend/main.py
lizlooney: Can you put a comment here that says under what circumstances this code is run?

lizlooney: Should the description just say 'Run the backend server' or 'Run the blocks backend server'?

frontend/storage/server_side_storage.ts
```diff
@@ -80,65 +85,67 @@ export class ServerSideStorage implements commonStorage.Storage {
...
 
   async list(path: string): Promise<string[]> {
-    // Ensure path ends with / for directory listing
-    const dirPath = path.endsWith('/') ? path : path + '/';
-    const response = await fetch(`${API_BASE_URL}/storage/${encodeURIComponent(dirPath)}`);
-    
+    if (!path.endsWith('/')) {
+      path += '/';
+    }
+    const response = await fetch(toStorageUrl(path));
```
lizlooney, line 91: I assume you can use encodeURIComponent because it will encode the / characters.
Is there any security risk in not encoding some characters? Like what if path contains `../`?

frontend/storage/server_side_storage.ts
```diff
@@ -80,65 +85,67 @@ export class ServerSideStorage implements commonStorage.Storage {
...
       throw new Error(`Failed to list files: ${response.statusText}`);
     }
-    
+
     const data = await response.json();
     return data.files || [];
   }
 
   async rename(oldPath: string, newPath: string): Promise<void> {
+    const stripped = (p: string) => p.startsWith('/') ? p.slice(1) : p;
```
lizlooney: Can you call this variable `strip` instead of `stripped`? Or even `stripLeadingSlash`?

lizlooney (changes requested):

frontend/storage/server_side_storage.ts
alan412: Yes, that is correct.  We don't want the / escaped.   We could do that and unescape it on the other end but because we are going inside of a database and not in the filesystem, it is still safe.

frontend/storage/server_side_storage.ts
alan412: Good idea.

lizlooney (approved):

PR #458, opened 2026-06-26
Back to snake case

lizlooney:
Back to snake case


frontend/storage/upgrade_project.ts
```diff
@@ -149,3 +151,17 @@ function noPreupgrade(moduleContentText: string): string {
 // @ts-expect-error: declared but not used
 function noUpgrade(_workspace: Blockly.Workspace): void {
 }
+
+async function upgradeFrom_0_1_x_to_0_2_0(
```
alan412, line 155: Should we have a function for upgrading when all it does is call a function?

frontend/storage/upgrade_project.ts
```diff
@@ -149,3 +151,17 @@ function noPreupgrade(moduleContentText: string): string {
...
+    projectName: string,
+    projectInfo: storageProject.ProjectInfo): Promise<void> {
+  // mrc_class_method_def blocks for mechanism 'opmodeStart' method need to be changed to 'opmode_start'.
+  // mrc_class_method_def blocks for mechanism 'opmodePeriodic' method need to be changed to 'opmode_periodic'.
+  // mrc_class_method_def blocks for mechanism 'opmodeEnd' method need to be changed to 'opmode_end'.
+  await upgradeBlocksFiles(
+      storage, projectName,
+      noModuleTypes, noPreupgrade,
+      isMechanism, upgrade_class_method_def_0_1_x_to_0_2_0);
+  projectInfo.version = '0.2.0';
```
alan412: This is redundant because at the end of all upgrading we set the version to the current version

alan412 (commented):
In general, this looks great.   Before we merge it we need to make sure that we have the updated robotpy and that the code that is generated still works.

frontend/storage/upgrade_project.ts
lizlooney: I like having a function because it gives us a place to put comments describing what is being changed.

frontend/storage/upgrade_project.ts
lizlooney: Done.

backend/deploy.py
```diff
@@ -71,11 +71,11 @@ def post(self) -> Response:
             with open(pyproject_path, "w", encoding="utf-8") as f:
                 toml_content = (
 f"""[tool.robotpy]
-robotpy_version = "2027.0.0.a6.post1"
+robotpy_version = "2027.0.0.a7"
```
alan412, line 74: Does this need to be 2027.0.0a7.post1

runtime_python/wpilib_blocks/color.py
```diff
@@ -134,6 +134,6 @@ def is_magenta(color: wpiutil.Color) -> bool:
     return False
 
 
-def get_ANSI(color: wpiutil.Color) -> str:
+def get_ansi(color: wpiutil.Color) -> str:
```
alan412, line 137: should ansi really be lowercase?  It refers to ANSI which correctly is uppercase.

alan412 (changes requested):
Mostly looks good.  Just 2 minor concerns

alan412:
I asked Claude code to review this PR and it had the following comments:
High — things that break at runtime or vanish from the UI
1. frontend/editor/extended_python_generator.ts:464 — the MECHANISM branch still injects when calls into 'opmodePeriodic', so it emits a dead camelCase method (no super call either) and mechanism when blocks silently stop firing.
2. runtime_python/wpilib_blocks/driver_station_display.py:10 — still calls color_utils.get_ANSI; every colored driver-station line raises AttributeError.
3. frontend/blocks/utils/python.ts:46 — ROBOT_METHOD_NAMES_OVERRIDEABLE still camelCase → "More Robot methods" toolbox section is empty.
4. frontend/blocks/utils/python.ts:60 — disabledPeriodic → the OpMode override block disappears.
5. frontend/types/hid_devices/*.ts (all 32 method: values) and frontend/fields/field_gamepads.ts:825 — every gamepad block generates a call that no longer exists in a7 (getSouthFaceButton, getLeftTriggerAxis, setRumble, …). Notably get_gamepad two lines above the rumble bug was fixed, so this is the same pass just stopping early.

Medium
6. python_tools/component.py:61 — sys never imported; the warning path raises NameError instead of printing (verified by running it), so a future robotpy arg rename aborts generation.
7. frontend/storage/upgrade_project.ts:62 — the 0.4.0 migration only covers mechanism opmode* method defs. Call blocks and method defs persist their Python names, so existing 0.3.1 projects still emit setThrottle(), def disabledPeriodic(self), etc. Either widen the rename table or document that old projects must be rebuilt.
8. python_tools/json_util.py:850 — get_component_args mutates and returns the shared list, so it isn't idempotent across overloads.

Low — stale rev.A301 comment plus a lost MotionCore default (json_util.py:110), and a dead empty-key placeholder row (json_util.py:306).

backend/deploy.py
lizlooney: Nope

alan412: never mind, I was wrong

runtime_python/wpilib_blocks/color.py
alan412: we decided that since robotpy uses ansi as lower case that we will follow

lizlooney: Lowercase is consistent with write_raw_ansi in wpilib.DriverStationDisplay.

alan412 (approved):
LGTM

PR #520, opened 2026-08-17
Display

alan412:
## Overview
Add DriverStationDisplay blocks for adding data and adding a line.   This also has a stub that simply prints right now until a7 is available from wpilib.

## Linked Issues
None.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that causes existing functionality to change)
- [ ] Documentation update
- [ ] Code style/formatting updates
- [ ] Refactoring / Renaming
- [ ] Other 

## How Has This Been Tested?
Created adding data block and adding line block, running code on systemcore and seeing them show up on the driver station console

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [N/A] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [N/A] I have run through the [docs/testing_before_pr_checklist.md]


frontend/blocks/mrc_driver_station_display.ts
lizlooney, line 1: Why not just use an MRC_CALL_PYTHON_FUNCTION block?

frontend/blocks/mrc_driver_station_display_line.ts
lizlooney, line 1: Why not just use an MRC_CALL_PYTHON_FUNCTION block?

frontend/blocks/mrc_driver_station_display.ts
alan412: Because the color has to be converted to ASCII escape codes.   We could do that in our wpilib_blocks I suppose.   (ie, move it out of TS into our python blocks).  Do you like that better?   Part of me does because it removes a special block.

frontend/blocks/mrc_driver_station_display_line.ts
alan412: same as above

frontend/blocks/mrc_driver_station_display.ts
lizlooney: yes, we should have python code that converts between ASCII escape code and wpi color.

I'm actually working on wpilib_blocks/color.py because I need/want to have functions that get hue from a color, etc, for the color range sensor. See https://github.com/lizlooney/systemcore-blocks-interface/blob/5c648c4c00e58fd9227885311c1ac10faded2263/runtime_python/wpilib_blocks/color.py

lizlooney: If you make the file `runtime_python/wpilib_blocks/color.py`, I'll just add my color functions there.

alan412: Ok.  I'll make a Display python that calls into color.py for the conversion and then sends the string to the wpilib one (and before it exists, it'll just print it without the escape codes.)

frontend/blocks/mrc_driver_station_display_line.ts
alan412: Now uses that block and does conversion in python.

frontend/blocks/mrc_driver_station_display.ts
alan412: Now uses that block and does conversion in python.

frontend/toolbox/driver_station_category.ts
```diff
@@ -21,34 +24,64 @@ export function getDriverStationCategory(editor: Editor): toolboxItems.Category
...
+}
+
+// Creates a real (non-shadow) block for the color input, plugged in as a
+// wpiutil.Color.WHITE class variable getter, so users can just change the
+// color using the dropdown rather than needing to find and attach the block
+// themselves.
+function createDefaultColorBlock(): {[key: string]: any} {
+    return {
+        block: {
+            type: MRC_GET_PYTHON_VARIABLE,
```
lizlooney: I think there is a function in mrc_get_python_variable.ts that create this kind of block. I think you have to pass it the variable data.

alan412: Good catch.    There was (but it wasn't exported).   I made the change.

frontend/blocks/mrc_get_python_variable.ts
```diff
@@ -518,7 +518,7 @@ function addModuleOrClassVariableBlocks(
   }
 }
 
-function createModuleOrClassVariableGetterBlock(
+export function createModuleOrClassVariableGetterBlock(
```
lizlooney: I'm sorry. I was on my phone when I suggested using this method and I didn't see all the arguments.

I've created https://github.com/alan412/systemcore-blocks-interface/pull/97 to add a new function that is more "friendly".

alan412: I merged it in.

lizlooney (approved):

PR #521, opened 2026-08-17
Add new custom renderer (compact Zelos)

alan412:
## Overview
This introduces a custom renderer that is like Zelos but not quite so humongous.   It also changes the default renderer to this new custom renderer.

## Linked Issues
None

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that causes existing functionality to change)
- [ ] Documentation update
- [ ] Code style/formatting updates
- [ ] Refactoring / Renaming
- [ ] Other 

## How Has This Been Tested?
Made sure that changing renderer worked, loaded a variety of workspaces to see what it looked like.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [N/A] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [N/A] I have run through the [docs/testing_before_pr_checklist.md]

### DIFF
--- a/frontend/reactComponents/ReadOnlyBlocklyPreview.tsx
+++ b/frontend/reactComponents/ReadOnlyBlocklyPreview.tsx
@@ -27,7 +27,6 @@ import 'blockly/blocks';
 import * as storageModuleContent from '../storage/module_content';
 import * as workspaces from '../blocks/utils/workspaces';
 import { themes } from '../themes/mrc_themes';
-import '../themes/compact_zelos'; // Registers the compact_zelos renderer.
 import { getGridConfig, getZoomConfig } from './BlocklyWorkspaceConfig';
 import { DEFAULT_ZOOM } from './UserSettingsProvider';
 
--- a/frontend/reactComponents/ReadOnlyBlocklyPreview.tsx
+++ b/frontend/reactComponents/ReadOnlyBlocklyPreview.tsx
@@ -27,6 +27,7 @@ import 'blockly/blocks';
 import * as storageModuleContent from '../storage/module_content';
 import * as workspaces from '../blocks/utils/workspaces';
 import { themes } from '../themes/mrc_themes';
+import '../themes/compact_zelos'; // Registers the compact_zelos renderer.
 import { getGridConfig, getZoomConfig } from './BlocklyWorkspaceConfig';
 import { DEFAULT_ZOOM } from './UserSettingsProvider';
 
--- a/frontend/reactComponents/UserSettingsProvider.tsx
+++ b/frontend/reactComponents/UserSettingsProvider.tsx
@@ -22,6 +22,7 @@
 
 import * as React from 'react';
 import { Storage } from '../storage/common_storage';
+import { COMPACT_ZELOS_RENDERER_NAME } from '../themes/compact_zelos';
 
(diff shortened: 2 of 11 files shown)


frontend/reactComponents/UserSettingsProvider.tsx
```diff
@@ -35,7 +36,7 @@ const USER_LAST_ZOOM_KEY = 'userLastZoom';
 export const DEFAULT_LANGUAGE = 'en';
 export const DEFAULT_THEME = 'dark';
 export const DEFAULT_SHOW_SIMPLE_CLASS_NAMES = true;
-export const DEFAULT_RENDERER = 'zelos';
+export const DEFAULT_RENDERER = COMPACT_ZELOS_RENDERER_NAME;
```
lizlooney: I thought you were going to make the default be Geras (Classic).

frontend/reactComponents/ReadOnlyBlocklyPreview.tsx
```diff
@@ -27,6 +27,7 @@ import 'blockly/blocks';
 import * as storageModuleContent from '../storage/module_content';
 import * as workspaces from '../blocks/utils/workspaces';
 import { themes } from '../themes/mrc_themes';
+import '../themes/compact_zelos'; // Registers the compact_zelos renderer.
```
lizlooney: This is already in frontend/reactComponents/BlocklyComponent.tsx 
Does it have to be here too?

frontend/reactComponents/UserSettingsProvider.tsx
alan412: I thought we started there and then decided if I could make a compact one like blocks but with the angle and circular ones then we would make it the default.

frontend/reactComponents/ReadOnlyBlocklyPreview.tsx
alan412: Good catch.   It isn't needed.  I just pushed a version that removes it.

frontend/reactComponents/UserSettingsProvider.tsx
alan412: If we are going to go to the effort to make a custom renderer, then I think we should make it look like what we think is best.   To do this, we should probably both use it so we can figure out what we don't like and change it.  It should probably be the default (which is what we'll use for all the documentation)

alan412:
Screenshot 2026-08-19 at 2 03 35 PM

alan412:
I changed the name (since it is now not simply a more compact Zelos) and I made it more like blocks but with the <> and () of Scratch.

Screenshot 2026-08-19 at 2 29 43 PM

frontend/reactComponents/UserSettingsProvider.tsx
lizlooney: Sorry, I hadn't put 2 (going back to Geras Classic) and 2 (changing the socket/plug for boolean) together.

Sounds good.

lizlooney (approved):
I'm approving this, but I will continue to use Geras style. Are you going to use FIRST style?

